PR #10, opened 2019-09-10
Fixes for guilding with gfortran

thomas-robinson:
Standards fixes to allow the model to be built with gfortran


thomas-robinson:
@bensonr please merge so I can update AM4 and CM4

PR #15, opened 2020-01-28
Updates needed for fms2_io

laurenchilutti:
These updates are needed to work with the FMS2 IO.

### DIFF
--- a/tools/fv_diagnostics.F90
+++ b/tools/fv_diagnostics.F90
@@ -25,7 +25,7 @@ module fv_diagnostics_mod
  use fms_mod,            only: write_version_number
  use fms_io_mod,         only: set_domain, nullify_domain, write_version_number
  use time_manager_mod,   only: time_type, get_date, get_time
- use mpp_domains_mod,    only: domain2d, mpp_update_domains, DGRID_NE
+ use mpp_domains_mod,    only: domain2d, mpp_update_domains, DGRID_NE, NORTH, EAST
  use diag_manager_mod,   only: diag_axis_init, register_diag_field, &
                                register_static_field, send_data, diag_grid_init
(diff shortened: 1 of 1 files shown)


tools/fv_diagnostics.F90
```diff
@@ -211,9 +211,9 @@ subroutine fv_diag_init(Atm, axes, Time, npx, npy, npz, p_ref)
 !                              set_name=trim(field), Domain2=Domain, tile_count=n)
 
        id_x = diag_axis_init('grid_x',grid_x,'degrees_E','x','cell corner longitude', &
-                           set_name=trim(field),Domain2=Atm(n)%Domain, tile_count=n)
+                           set_name=trim(field),Domain2=Atm(n)%Domain, tile_count=n, pos=EAST)
```
thomas-robinson: Change this from `pos` to `domain_position`

tools/fv_diagnostics.F90
```diff
@@ -211,9 +211,9 @@ subroutine fv_diag_init(Atm, axes, Time, npx, npy, npz, p_ref)
...
        id_y = diag_axis_init('grid_y',grid_y,'degrees_N','y','cell corner latitude',  &
-                           set_name=trim(field), Domain2=Atm(n)%Domain, tile_count=n)
+                           set_name=trim(field), Domain2=Atm(n)%Domain, tile_count=n, pos=NORTH)
```
thomas-robinson: Change this one also to `domain_position`

thomas-robinson:
This one is needed to be compatible with NOAA-GFDL/FMS#270 and the tag 2020.01.  It should be done before NOAA-GFDL/FMS#270 is accepted so that we can test with this code.

PR #21, opened 2020-04-14
Bug fix for two-way nest updating

lharris4:


model/fv_nesting.F90
```diff
@@ -2573,14 +2447,18 @@ subroutine twoway_nest_update(npx, npy, npz, zvir, ncnst, sphum,     &
               neststruct%isu, neststruct%ieu, neststruct%jsu, neststruct%jeu, &
               npx, npy, npz, 0, 0, &
               neststruct%refinement, neststruct%nestupdate, upoff, 0, &
-              neststruct%parent_proc, neststruct%child_proc, parent_grid, grid_number-1)
+              parent_grid%neststruct%parent_proc, neststruct%child_proc, parent_grid, grid_number-1)
 
       endif !conv_theta
 
       call mpp_sync!self
```
bensonr, line 2454: Do you want to perform an mpp_sync_self or is the full sync necessary here and down below the other update_coarse_grid?

lharris4:
Hi, Rusty. What is the difference between mpp_sync_self() and mpp_sync()?

Thanks,
Lucas

bensonr:
@lharris4 _mpp_sync_ requires all MPI-ranks to check in and wait for all to arrive before proceeding.  _mpp_sync_self_ only ensures that a give MPI-rank's outstanding sends and receives are completed before proceeding.

lharris4:
Hi, Rusty. I usually use mpp_sync(). I don't recall if I had problems using
mpp_sync_self().

Thanks,
Lucas

bensonr:
@lharris4 No worries - we'll publish it as is and explore the alternatives later.

Issue #38, opened 2020-04-30
Possible bug in fv_grid_tools.f90


ofuhrer:
The code [here](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/75b6ba3519dc91092ca8bcf409493a4182fa5712/tools/fv_grid_tools.F90#L620) on L620-621 in fv_grid_tools.f90 is...
`
if (ABS(grid_global(i,j,1,1)) < 1.d-10) grid_global(i,j,1,1) = 0.0
if (ABS(grid_global(i,j,2,1)) < 1.d-10) grid_global(i,j,2,1) = 0.0
`
...and should probably be...
`
if (ABS(grid_global(i,j,1,n)) < 1.d-10) grid_global(i,j,1,n) = 0.0
if (ABS(grid_global(i,j,2,n)) < 1.d-10) grid_global(i,j,2,n) = 0.0
`
...or taken out of the loop over regions.

lharris4:
Hi, Oliver. Indeed the last index should be n.

Thanks,
Lucas

bensonr:
@ofuhrer @lharris4 please make this fix and put in a PR for master.

ofuhrer:
See [PR39](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/39/files).

bensonr:
@ofuhrer I've assigned @lharris4 and @bensonr as reviewers

PR #39, opened 2020-04-30
Fix mesh generation in init_grid()

ofuhrer:
Tile index to `grid_global` was not correct inside a loop over tiles.

@bensonr @lharris4 This fixes [Issue 38](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/38).

### DIFF
--- a/tools/fv_grid_tools.F90
+++ b/tools/fv_grid_tools.F90
@@ -617,8 +617,8 @@ contains
 !----------------------------------------------------------------------------------------------------
                       if ( grid_global(i,j,1,n) < 0. )              &
                            grid_global(i,j,1,n) = grid_global(i,j,1,n) + 2.*pi
-                      if (ABS(grid_global(i,j,1,1)) < 1.d-10) grid_global(i,j,1,1) = 0.0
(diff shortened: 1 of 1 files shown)


bensonr:
@lharris4 will this fix be answer changing and under what conditions?

ofuhrer:
My 0.02$: This will change the resulting lat/lon of the mesh, if mesh generation is done internally in the model. I am assuming that typically a INPUT/grid_spec.nc file is being read for mesh generation and mesh generation is only done for idealized model setups, but I might be wrong. Also, it could be that the next time somebody generates a grid_spec.nc file using the model after the fix, that the contents of that file will be different.

thomas-robinson:
This will affect the doubly periodic and aquaplanet then

lharris4:
This would indeed change answers, but only if the grid is generated on-line
and any of the values are within the range (-1.d-10,1.d-10) and not
identically zero. In the unrotated cube this should not a problem but
someone should check.

Lucas

lharris4:
#39 should fix this

Issue #41, opened 2020-05-06
Read before write of sw_corner et al. in init_grid()?


ofuhrer:
The code in `fv_control_init()` in `fv_control.F90` loops over all grids and calls `init_grid()` on L590.
https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/eee655598d58836a600c52a849e2ed69ab910653/model/fv_control.F90#L589-L596
Right after that grid_utils_init() is being called on L596. It turns out that that `sw_corner`, `se_corner`, `ne_corner` and `nw_corner` are only being set in the latter on L218.
https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/eee655598d58836a600c52a849e2ed69ab910653/model/fv_grid_utils.F90#L218-L228
But these are being read by the former (see L888 and following to compute `area_c`).
https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/eee655598d58836a600c52a849e2ed69ab910653/tools/fv_grid_tools.F90#L888-L897
This seems like a bug.

Note: Reported by @rheacangeo

lharris4:
Hi, Oliver. This only affects the creation of area_c and rarea_c, and
apparently does not make more than minimal changes. You can move the
assignment of ??_corner to the routine which is called earlier to ensure
correctness of these two arrays at the corners.

Lucas

lharris4:
Hi, Oliver. I would prefer it not to be moved into fv_arrays.F90 since this
routine is purely for allocating memory.

Lucas

bensonr:
PR #113 cherry-picked into master branch

PR #42, opened 2020-05-07
Fix initialization of ??_corner flags

ofuhrer:
This fixes issue https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/41 where of the `??_corner` flags in `init_grid()` in `tools/fv_grid_tools.F90` are being read before setting them. The initialization of `??_corner` has been moved to `tools/fv_grid_tools.F90`.

Side note to @bensonr and @lharris4: It would be great to have some regression tests integrated into GitHub in order to make sure that such PR's actually pass regression.

Thanks to @rheacangeo for the bug report!

### DIFF
--- a/model/fv_grid_utils.F90
+++ b/model/fv_grid_utils.F90
@@ -215,18 +215,6 @@
       cos_sg(:,:,:) =  big_number
       sin_sg(:,:,:) = tiny_number
 
-      sw_corner = .false.
-      se_corner = .false.
-      ne_corner = .false.
-      nw_corner = .false.
-
-      if (grid_type < 3 .and. .not. Atm%gridstruct%bounded_domain) then
-         if (       is==1 .and.  js==1 )      sw_corner = .true.
-         if ( (ie+1)==npx .and.  js==1 )      se_corner = .true.
-         if ( (ie+1)==npx .and. (je+1)==npy ) ne_corner = .true.
-         if (       is==1 .and. (je+1)==npy ) nw_corner = .true.
-      endif
-
   if ( sw_corner ) then
        tmp1 = great_circle_dist(grid(1,1,1:2), agrid(1,1,1:2))
        tmp2 = great_circle_dist(grid(1,1,1:2), agrid(2,2,1:2))
--- a/tools/fv_grid_tools.F90
+++ b/tools/fv_grid_tools.F90
@@ -565,6 +565,18 @@ contains
     latlon = .false.
     cubed_sphere = .false.
 
+    ! set corner flags
+    sw_corner = .false. 
+    se_corner = .false. 
+    ne_corner = .false. 
+    nw_corner = .false. 
+    if (Atm%flagstruct%grid_type < 3 .and. .not. Atm%gridstruct%bounded_domain) then 
+       if (       is==1 .and.  js==1 )      sw_corner = .true. 
+       if ( (ie+1)==npx .and.  js==1 )      se_corner = .true. 
+       if ( (ie+1)==npx .and. (je+1)==npy ) ne_corner = .true. 
+       if (       is==1 .and. (je+1)==npy ) nw_corner = .true. 
+    endif  
+
     if ( (Atm%flagstruct%do_schmidt .or. Atm%flagstruct%do_cube_transform) .and. abs(atm%flagstruct%stretch_fac-1.) > 1.E-5 ) then
        stretched_grid = .true.
        if (Atm%flagstruct%do_schmidt .and. Atm%flagstruct%do_cube_transform) then


lharris4:
Hi, Oliver. Thanks.

We are working on a set of regression tests but I don't think there is a
good way to run them on GitHub. I know the MOM6 team has a way to do it but
it took them years to set it all up.

Lucas

underwoo:
> Hi, Oliver. Thanks. We are working on a set of regression tests but I don't think there is a good way to run them on GitHub. I know the MOM6 team has a way to do it but it took them years to set it all up. Lucas

@lharris4, the MOM6 team has two types of tests.  The small tests (mainly a few builds with small tests to verify the executable works and reproduces) are run on [Travis CI](https://travis-ci.com/github/underwoo/MOM6).  After the Travis CI  tests pass, they bring the updates onto our internal GitLab instance where they run longer tests on Gaea.  If you would like to look into doing this with GFDL_atmos_cubed_sphere, we can talk with the MOM group and help get you setup.  However, outside collaborators will not be able to see the results of these tests.

We are also in the process of investigating the use of one or more of the cloud providers (e.g. AWS, Google Compute, Azure, etc.)  to run full regression tests.  Last year, @thomas-robinson and myself showed it was possible to run regression tests on one of the cloud providers.  With the help of others, we are working to acquire funding to implement the full solution.  This will allow us to use [GitHub Actions](https://github.com/features/actions) configured to use the cloud provider to run both quick verification (build, style, etc) tests, and when desired full regression tests.

If you would like more information on this, we can put a meeting together to answer any questions you and others may have.

ofuhrer:
Thanks @lharris4 and @underwoo for your answers.

We've used GitHub integration for TravisCI, CircleCI and Jenkins and it works great. The advantage of making test results also visible to contributors who make a PR, is that you can push the responsibility of getting tests to pass to the developers which takes a big burden off the shoulders of the code maintainers. This is how a PR would look like...

image

At least for [MOM5](https://github.com/mom-ocean/MOM5) it seems that the TravisCI [test results ](https://travis-ci.org/github/mom-ocean/MOM5/builds/560844582) are visible to outside users.

We would be more than happy to help with this as well, if there's anything you can see us do.

nbren12:
I would just add that even a simple test usingTravis/github/CircleCIs free tier would be an incredibly useful tool. This practice is ubiquitous in the open source community for a reason, it greatly amplifies the trust between maintainers and contributors, and can greatly accelerate development.

The great thing about CI systems is you can add complexity incrementally. At Vulcan, we started with bit-for-bit regression tests of the C48 model (which took approx 2 week to set up, including learning how to build the model in docker), and now we have much more sophisticated system, but that doesn't need to happen overnight.

bensonr:
Thanks @nbren12 @ofuhrer for your suggestions.

We use a combination of Travis and GH actions for FMS CI today and are building containers for a tier of real data tests with GFDL models.  We have a plan to use these same concepts to build a CI system for the dycore.  Anything you can offer to accelerate the process will be appreciated.

I'd like to move this conversation to issue #43 with collaboration occurring via GH, direct email, and video conferences as needed.

bensonr:
@nbren12 can you perform a code review on the updates in this PR - thanks

nbren12:
@bensonr I wish I could. I’m not really an expert on this part of the code base, just got pulled in for the CI discussion. Maybe @rheacangeo can take a look?

rheacangeo:
The code changes look good to me, but it does change 'the answer', so it's a question of whether that change is acceptable. Also whether it causes other consequences. e.g. there is a bunch of edge and corner handling of area_c after the code block that currently isn't executed in master because the corner variables aren't set yet. It doesn't look like any of them have the same impact, but it's possible for something there to be 'compensating' for the corner code not getting executed. But more likely, I would guess this routine was written assuming the corner variables had already been set and it's just a bug fix . In our regression tests using C12, this changes files:
atmos_dt_atmos.tile*.nc
sfc_dt_atmos.tile*.nc
and all of the tile restart files
RESTART/*.tile*.nc
All the variables I looked at small, but not negligible differences. e.g. I attached difference plots of  UGRD at 500mb (~0.4 m/s) and RH at 850m (~0.2%) differences. 
U_500_diff
RH_850_diff

But with such a small number of cells the impacts are likely exaggerated. It might make sense to check out diffs on a more realistic resolution.

ofuhrer:
Not sure where this is at. I guess the change does cause meteorological differences and thus should go through some sort of testing on GFDL side.

lharris4:
Hi, Oliver. Thanks for bringing this to our attention. This probably fell
through the cracks with all that was going on in May. I should take a look.

Lucas

bensonr:
Lucas - let me know when you've finished your analysis by performing a formal review.  Once done, I will merge it in.

lharris4:
Hi, all. I have noticed some unexpected changes in the way that two grid
coefficients are computed with this change:

after: 735: da_max_c/da_min_c=   3.04909832895044
before: 735: da_max_c/da_min_c=   2.28672617943394

which appear to be due to changes in how the areas of the corner cells of
the dual grid (which are properly hexagons and not squares) are computed.

This change also results in a change to how the artificial diffusion is
applied globally, and may lead to significant changes in the solution.

**However** I think that after applying this fix this is the *correct*
behavior: with this fix the areas of the dual-grid cells in the corners are
correct. I will do some tests to check.

Thanks,
Lucas

lharris4:
Hi, all. Upon further examination I believe that there is a lot more that
would need to be changed to implement this than just moving the
declarations for ??_corner to init_grid(). So let's leave out the code
change for now until this can be more thoroughly considered.

Thanks,
Lucas

ofuhrer:
Hi Lucas,
All good, just note that in the current implementation these flags are being read before they are actually being set (see original Issue https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/41 .

lharris4:
Hi, Oli. Thanks for the reminder. In that case you can simply set the four
flags to .false. in fv_arrays, where they are defined.

Lucas

bensonr:
Fixed via PR #113 and cherry-picked to master.

PR #47, opened 2020-08-27
User/wfc/climate nudging

wfcooke:
Updates for PV350K, PV550K diagnostics and some climate nudging updates our division needs.

### DIFF
--- a/model/fv_arrays.F90
+++ b/model/fv_arrays.F90
@@ -60,6 +60,9 @@ module fv_arrays_mod
            id_dbz, id_maxdbz, id_basedbz, id_dbz4km, id_dbztop, id_dbz_m10C, &
            id_ctz, id_w1km, id_wmaxup, id_wmaxdn, id_cape, id_cin
 
+! Selected theta-level fields from 3D variables:
+ integer :: id_pv350K, id_pv550K
+
 ! Selected p-level fields from 3D variables:
  integer :: id_vort200, id_vort500, id_w500, id_w700
  integer :: id_vort850, id_w850, id_x850, id_srh25, &
--- a/tools/fv_diagnostics.F90
+++ b/tools/fv_diagnostics.F90
@@ -708,6 +708,10 @@ contains
           !--------------------
           idiag%id_pv = register_diag_field ( trim(field), 'pv', axes(1:3), Time,       &
                'potential vorticity', '1/s', missing_value=missing_value )
+          idiag%id_pv350K = register_diag_field ( trim(field), 'pv350K', axes(1:2), Time,       &
+                           '350-K potential vorticity; needs x350 scaling', '(K m**2) / (kg s)', missing_value=missing_value)
+          idiag%id_pv550K = register_diag_field ( trim(field), 'pv550K', axes(1:2), Time,       &
+                           '550-K potential vorticity; needs x550 scaling', '(K m**2) / (kg s)', missing_value=missing_value)
 
           ! -------------------
           ! Vertical flux correlation terms (good for averages)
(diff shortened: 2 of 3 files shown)


lharris4:
Hi, Rusty. Thanks. Perhaps one of the regression tests should be a nudging
test? It would not have to be particularly high-resolution, c48 or c192
would be fine.

Lucas

laurenchilutti:
This change has been tested in SHiELD and the results are consistent.

thomas-robinson:
@uramirez8707 do you have a fix for this crash?

uramirez8707:
This is all that is needed: 
https://github.com/uramirez8707/GFDL_atmos_cubed_sphere/commit/86143aa6b77ac9d0a8302c71c33b2680e425685f

The crash happens for File=`19210101.atmos_daily.ens_03.tile2.nc.0003` Field=`z925`

This does **not** crash when using FMS-2019.01.02 + GFDL_atmos_cubed_sphere - 2020.01
This does crash when using FMS-2019.01.03 (nest updates) + GFDL_atmos_cubed_sphere - 2020.02 
This does crash when using FMS-2020.02 (nest updates + FMS2io in diag_manager) + GFDL_atmos_cubed_sphere - 2020.02

One of the changes between GFDL_atmos_cubed_sphere  2020.01 and 2020.02 is the removal of the `ngc` argument in the `get_height_given_pressure` subroutine. 

https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/72bbfcc67d5a119038f2a77ea3a4f8d752141017/tools/fv_diagnostics.F90#L1557

https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/2aac011c5a591e87070ff6d9411a3ba21fdef945/tools/fv_diagnostics.F90#L2047

The `get_height_given_pressure` subroutine is where `a3` is defined so it is probably not defined for all the grid points. 

Initiliazing `a3` to 0 avoids the crash, but I do not know if that is the *correct* thing to do.

laurenchilutti:
@lharris4 @bensonr Is there a reason that lines 3072-3076 from the [```get_height_given_pressure```](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/19b48bc1f3c6a86ff7eb73fc9c219e8020220c58/tools/fv_diagnostics.F90#L3045) subroutine were removed in commit f44db53248c745daaa80e4c85cc534881eff6b37? I think that if we added this back into the subroutine, ```a3``` wouldn't need to be initialized to 0 as Uriel suggests.  I have not tested this though, so I could be wrong.

https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/19b48bc1f3c6a86ff7eb73fc9c219e8020220c58/tools/fv_diagnostics.F90#L3072-L3076

bensonr:
@laurenchilutti I don't think we want to reinstate computation for pressures that may be below the surface of the earth.  I agree the height should be either missing value or 0.

bensonr:
@wfcooke Please see the comment from @uramirez8707 and @thomas-robinson regarding crashes due to bad values for a3 from the get_height_given_pressure computation.  As @laurenchilutti points out, the calculation may abort early in high altitude areas.  Decide if your calculation should produce zeros or missing-value for those areas or missing_value and update the code accordingly.  Once you have done this, the updates can be merged.

wfcooke:
Rusty, Uriel, Tom
I talked to the folks in my division about this update and they think a
missing value is probably the best way to go.
The segment of code you pointed to was not a place that I checked in a
modiffication for.
If I read the "blame" field properly it seems to be related to EMC (checked
in by Rusty?). I don't want to mess them up.

Can you guys change that part of the code or do you need me to add a
a2(i,j,n) =missing_value
before the 1000 continue line?

Will

bensonr:
@wfcooke the removal was not done by or for EMC.  Please make the update
you suggested so we can get this merged into the dev/gfdl branch.

wfcooke:
@bensonr  : Sorry about that. I missed an email.
Code is updated, commit should be 4dacd7e8fd0c485d32622734685c56ad059baa65 from my end.

PR #62, opened 2020-12-02
ufs-release/public-v2: initialized cld_amt to zero for regional runs

climbfuji:
## Description

This PR contains a one-liner change in `model/fv_regional_bc.F90` to initialize `cld_amt` to zero so that regional runs using GFDL microphysics can be initialized from input data that doesn't contain `cld_amt` (e.g. RAP/HRRR input data).

## Testing

See https://github.com/ufs-community/ufs-weather-model/pull/313

## Dependencies

https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/62
https://github.com/NOAA-EMC/fv3atm/pull/207
https://github.com/ufs-community/ufs-weather-model/pull/313

### DIFF
--- a/model/fv_regional_bc.F90
+++ b/model/fv_regional_bc.F90
@@ -3701,7 +3701,7 @@ subroutine remap_scalar_nggps_regional_bc(Atm                         &
 ! If the source is from old GFS or operational GSM then the tracers will be fixed in the boundaries
 ! and may not provide a very good result
 ! 
-!  if (cld_amt .gt. 0) BC_side%q_BC(:,:,:,cld_amt) = 0.
+  if (cld_amt .gt. 0) BC_side%q_BC(:,:,:,cld_amt) = 0.
   if (trim(data_source) /= 'FV3GFS GAUSSIAN NEMSIO FILE') then
    if ( Atm%flagstruct%nwat .eq. 6 ) then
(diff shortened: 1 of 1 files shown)


climbfuji:
@bensonr Hopefully the second-to-last PR for the ufs-release/public-v2 branch. This change may have to find its way back to dev/emc, because the scenario to initialize regional runs using GFDL MP from RAP/HRRR data is a quite likely.

bensonr:
@climbfuji - do you know when and what the last PR is expected to be?

climbfuji:
I'd love to know ... we are seeing b4b differences when changing the MPI decomposition in regional runs, while in global runs the same set of physics gives b4b identical results when changing the MPI decomposition.

I am not sure yet where the problem is, it could be in the regional init code in the dycore, but it could also be in fv3atm or in the physics (although the latter is unlikely, because the physics are column based, but who knows). What I do know is that the code freeze is around mid/end of January, though.

climbfuji:
This PR is ready to merge, all regression tests passed (see https://github.com/ufs-community/ufs-weather-model/pull/313).

bensonr (approved):

climbfuji:
Thanks, @bensonr !

PR #65, opened 2021-01-15
Merge NOAA-EMC dev/emc

XiaqiongZhou-NOAA:
Merge NOAA-EMC dev/emc


lharris4:
Hi, all. After consultation with @Xi Chen - NOAA Affiliate
<xi.chen@noaa.gov> and some further thought and experimentation I am now
very suspicious of one of the changes being made in nh_core.F90. In
particular this change of the code:
do k=km, 1, -1
do i=...
gz(i,j,k) = max( gz(i,j,k), gz(i,j,k+1) + dz_min )
enddo
enddo

To:
do k=2, km+1
do i=...
gz(i,j,k) = min( gz(i,j,k), gz(i,j,k-1) - dz_min )
enddo
enddo
is questionable since this has much larger and non-local effects than the
original version does. In the original version S-J Lin wrote, noting that
in general the difference in gz (equal to -delz) gets smaller with
increasing k, the fixer acts by raising layer interface k so that it is at
least dz_min above interface k+1. This effectively acts by borrowing height
from the layer directly above it, without altering layers further up in the
atmosphere---that is, at smaller k indices (since k=1 is the top of the
domain in FV3). After fixing up this level, the fixer keeps iterating
downward, again fixing as necessary.

However, the new code acts by borrowing from layer level below it. This
makes this layer, which is nearly always thinner than the layer above
interface k, thinner still, and makes it much more likely that this layer
is also in need of fixing. This could then keep going until it reaches
k=km+1, which is the height of the topography. The fixer would then
artificially lower the topography to ensure that the lowest layer is at
least dz_min thick.

While this never feeds back on the actual topography array phis (and the
layer km+1 of the height array is overwritten by the static topography at
the end of Riem_Solver3), it could potentially lead to inconsistency
between the real surface and the computed surface in the semi-implicit
solver, leading to noise and/or inaccurate solutions. Also, when the
updated zh is computed at the end of Riem_solver3, the shift in dz from the
artifically lowered topography will lead to a lifting of the entire column
when those dzs are then used with the correct topography to reconstruct gz,
which is then passed into the horizontal pressure gradient force routine.
With the original method, since the topography is not changed and only one
interface is shifted, this does not happen.

The change to dz_min is OK for the time being and our own testing has shown
no problems. But as I indicated earlier this is intended only as a quick
fix to meet operational demands on a short time constraint, and there is a
serious need for the root problem for the crashes in GFSv16 and RRFS to be
found and fixed.

Lucas

XiaqiongZhou-NOAA:
Hi, Lucas, Xi Chen:
Thank you very much for carefully considering the code change. That is great. 

I put some figures during the tests, here is the link 
https://docs.google.com/presentation/d/1A4_YfkVNn47Fc1AyJTijUBBhQ4s-EMCwxjHcgkulO8k/edit?usp=sharing

Yes, the difference between these two formula is the the Lin's original formula is to adjust the height 
above, but the revised one is to adjust the height below. My original thought was based on the slide 1 as the small delz 
is likely from the lowest level (z (km+1), black line), so it is better to adjust the lower interface. The slide 2 and 3 is the comparison after dz_min increasing to 4 with the original formula and revised formula so that the model will not crash. I think they both cause some issues in a short time period. I can not tell which is better. I am OK to change them back to Lin's original one. I agree it is better to find the root reason to fix the issue.

yangfanglin:
Based on more offline discussions between EMC and GFDL, I'd suggest 

1) still use the original equation 

do k=km, 1, -1
do i=...
gz(i,j,k) = max( gz(i,j,k), gz(i,j,k+1) + dz_min )
enddo
enddo

set default dz_min to its original value

2) include zero-gradient BC fix as another namelist option

3) For RRFS, GFS.v17/GEFS.v13 development,  option #2 shall be exercised.

4) Keep looking for the cause of  model crashes;  test other solutions, including applying different terrain filters.

Fanglin

junwang-noaa:
Kate, do you want to add code changes from your [PR#53](https://github.com/NOAA-EMC/GFDL_atmos_cubed_sphere/pull/53)?

bensonr (approved):

XiaqiongZhou-NOAA:
WangJun:
What kind of change are you referring to?
I have updated the code yesterday as Fanglin suggested based on
Lucus's concern from the previous update with dz_min. Nothing else in my
mind.
Best
Kate

bensonr:
@junwang-noaa @XiaqiongZhou-NOAA - Let's finalize and merge this PR first.  Once this is completed, we can then create  new PR for the updates currently pending to the NOAA-EMC fork in PR [#53](https://github.com/NOAA-EMC/GFDL_atmos_cubed_sphere/pull/53)

XiaqiongZhou-NOAA:
Ok, I see. It is for the merge to GFDL.
I agree. I can create new PR later.

junwang-noaa (approved):

laurenchilutti (approved):

PR #68, opened 2021-02-05
ufs-release/public-v2: bit-for-bit reproducibility for regional runs when changing MPI decomposition

climbfuji:
This PR fixes b4b mismatches when changing the MPI decomposition for regional runs (see https://github.com/ufs-community/ufs-weather-model/pull/409 and https://github.com/ufs-community/ufs-weather-model/issues/288) for more details.

- add missing call to `mpp_update_domain` for `q` in `model/fv_tracer2d.F90`
- add missing calls and update existing calls to update domain and halos, fix compiler warnings about additional text after `#endif` statements in `model/dyn_core.F90`
- remove two unused variables from `model/sw_core.F90`
- fix a few trailing whitespaces

**Note 1.** Each of the added/modified calls to update the domain or halos is required. After achieving b4b identical results, I removed each of the changes shown here and left the rest, and every time the results were no longer b4b.

**Note 2.** The perhaps most confusing change is to move the call to `call start_group_halo_update(i_pack(8), u, v, domain, gridtype=DGRID_NE)` to further down in the code in `model/sw_core.F90`, because there is a call to a regional boundary update for `u` and `v` after the original location.

**Note 3.** I am not an expert in the FV3 dycore, and it is possible that there are alternative and easier ways to achieve the same.

Associated PRs:

https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/68
https://github.com/NOAA-EMC/fv3atm/pull/245
https://github.com/ufs-community/ufs-weather-model/pull/409

For regression testing, see https://github.com/ufs-community/ufs-weather-model/pull/409.

### DIFF
--- a/model/fv_tracer2d.F90
+++ b/model/fv_tracer2d.F90
@@ -749,6 +749,9 @@ subroutine tracer_2d_nested(q, dp1, mfx, mfy, cx, cy, gridstruct, bd, domain, np
                                                reg_bc_update_time,      &
                                                iq )
             enddo
+            call timing_on('COMM_TOTAL')
+            call mpp_update_domains(q, domain, complete=.true.)
+            call timing_off('COMM_TOTAL')
       endif
 
 
--- a/model/dyn_core.F90
+++ b/model/dyn_core.F90
@@ -555,19 +555,23 @@ contains
       endif
 
       if (flagstruct%regional) then
+        call timing_on('COMM_TOTAL')
         reg_bc_update_time=current_time_in_seconds+bdt*(n_map-1)+(0.5+(it-1))*dt
         call regional_boundary_update(delpc, 'delp', &
                                       isd, ied, jsd, jed, npz, &
                                       is,  ie,  js,  je,       &
                                       isd, ied, jsd, jed,      &
                                       reg_bc_update_time )
+        call mpp_update_domains(delpc, domain, complete=.true.)
 #ifndef SW_DYNAMICS
         call regional_boundary_update(ptc, 'pt', &
                                       isd, ied, jsd, jed, npz, &
                                       is,  ie,  js,  je,       &
                                       isd, ied, jsd, jed,      &
                                       reg_bc_update_time )
+        call mpp_update_domains(ptc, domain, complete=.true.)
 #endif
+        call timing_off('COMM_TOTAL')
       endif
 
       if ( hydrostatic ) then
@@ -670,7 +674,7 @@ contains
 
            endif
 
-#endif SW_DYNAMICS
+#endif
 
       endif   ! end hydro check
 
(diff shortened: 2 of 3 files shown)


climbfuji:
@junwang-noaa @jamesabeles-noaa @dusanjovic-noaa

climbfuji:
@bensonr

XiaqiongZhou-NOAA (commented):
Reviewed. It looks good to me.

model/fv_tracer2d.F90
```diff
@@ -749,6 +749,9 @@ subroutine tracer_2d_nested(q, dp1, mfx, mfy, cx, cy, gridstruct, bd, domain, np
                                                reg_bc_update_time,      &
                                                iq )
             enddo
+            call timing_on('COMM_TOTAL')
```
lharris4, line 752: This is OK but could likely be avoided by having the regional boundary update fill boundary rows that lie in the haloes too: ie. if this is a west edge, then have the BCs fill in (-2:0) x (jsd:jed) instead of just (-2:0) x (jsc:jec).

lharris4: I think the nesting code already does this.

climbfuji: @lharris4 Can you explain please what you mean? I tried to skip this one update call, and the results were different. I did that with every single change in this PR, actually. Took them out on by one  but left the others and reran the tests. Every time they failed, thus I think that all of them are required (unless changes are made to the regional BC update code as you mentioned).

lharris4:
Hello. That is exactly what I mean: I think that if the regional BC update
code is changed, this mpp_update_domains call is unnecessary.

Lucas

model/fv_tracer2d.F90
climbfuji: > Hello. That is exactly what I mean: I think that if the regional BC update code is changed, this mpp_update_domains call is unnecessary. Lucas
> […](#)

Got it, thank you. Do you think it is ok to get the changes into the public release code as they are now, or should we try to follow your suggestion. The reason I am asking is that the SRW App is on a tight release schedule. The code freeze was last week and we were hoping to get the b4b issues fixed and merged as an "emergency bugfix".

We should definitely follow your suggestions for the authoritative/development branches and not try to bring over the public release code changes. Doing so would probably allow us to remove several of the already existing calls to `mpp_update_domain`, if I understand this correctly, but it will require a bit of time for the development to happen.

lharris4:
Hi, Dom. Sure, this is fine.

Lucas

climbfuji:
> Hi, Dom. Sure, this is fine. Lucas
> […](#)

Thanks, Lucas. I'll post another note when all the regression tests have passed and we are ready to merge.

climbfuji:
@lharris4 this is ready to merge whenever it is convenient for you - thanks!

bensonr:
With @lharris4 okaying this PR, I'll merge it now.

model/fv_tracer2d.F90
junwang-noaa: @lharris4 @bensonr I would like to follow up with this issue, do we have the boundary fill updates in the master and dev/emc branch? We still have decomposition reproducibility issue with regional test and would like to see if we can try the fix you mentioned.

lharris4:
Hi, Jun. We will merge these changes into master and release them alongside
our updates for the regional nesting and some other bugfixes.

Thanks,
Lucas

junwang-noaa:
That's great, thanks for letting me know!

PR #69, opened 2021-02-10
Fix for GNU compiler

kaiyuan-cheng:
1. Revision to ensure compatibility with the GNU compiler.
2. Fix for the regional domain where the model failed to initialize the grid.


lharris4 (approved):
Looks good, and the explicit loops are something I had been hoping to have for a while.

laurenchilutti (approved):

PR #73, opened 2021-03-04
Feb 2021: add dz_min as namelist variable and add an option to use zero-gradient BC to reconstruct interface u/v with PSM

junwang-noaa:
There is one commit ([PR#53](https://github.com/NOAA-EMC/GFDL_atmos_cubed_sphere/pull/53)) from EMC fork:

* 1)Add an option to use original BC or zero-gradient BC to reconstruct interface u/v with PSM for height advection
2)Change dz_min as a namelist input
* Change to the original version by adjust the upper interface of ZH when dz<dz_min
* add authors for the revision


bensonr:
This has already been included via other mechanisms.

PR #76, opened 2021-03-16
changes dz_min from defined constant to a namelist option

bensonr:
Second step of the synchronization process for dev/emc branch within the NOAA-GFDL organization.

### DIFF
--- a/model/nh_utils.F90
+++ b/model/nh_utils.F90
@@ -196,13 +196,13 @@ CONTAINS
 ! Enforce monotonicity of height to prevent blowup
 !$OMP parallel do default(none) shared(is1,ie1,js1,je1,ws,zs,gz,rdt,km,dz_min)
   do j=js1, je1
-     do k=2, km+1
-        do i=is1, ie1
-           gz(i,j,k) = min( gz(i,j,k), gz(i,j,k-1) - dz_min )
-        enddo
-     enddo
      do i=is1, ie1
         ws(i,j) = ( zs(i,j) - gz(i,j,km+1) ) * rdt
+     enddo
+     do k=km, 1, -1
+        do i=is1, ie1
+           gz(i,j,k) = max( gz(i,j,k), gz(i,j,k+1) + dz_min )
+        enddo
      enddo
   enddo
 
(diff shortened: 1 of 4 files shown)


XiaqiongZhou-NOAA (approved):
It looks fine to me.
Just wondering why not merge the EMC fork to GFDL emc/dev since these are already included in the EMC fork. 
Should the description be added in fv_control.F90 as before ? " !> \param[in] dz\_min  Real: Minimum layer thickness to enforce monotonicity of height to prevent blowup. Set to 2 by default."

laurenchilutti (approved):

model/nh_utils.F90
```diff
@@ -195,7 +194,7 @@ subroutine update_dz_c(is, ie, js, je, km, ng, dt, dp0, zs, area, ut, vt, gz, ws
 6000 continue
 
 ! Enforce monotonicity of height to prevent blowup
-!$OMP parallel do default(none) shared(is1,ie1,js1,je1,ws,zs,gz,rdt,km)
+!$OMP parallel do default(none) shared(is1,ie1,js1,je1,ws,zs,gz,rdt,km,dz_min)
   do j=js1, je1
      do k=2, km+1
```
junwang-noaa, line 199: This is different from current EMC fork, where I think Kate switched back to k=km,1,-1 in her [PR#53](https://github.com/NOAA-EMC/GFDL_atmos_cubed_sphere/pull/53)
https://github.com/NOAA-EMC/GFDL_atmos_cubed_sphere/pull/53/files#diff-ab35d906facb07c24877b72bbc6b6acc3e856a82cb065a5a6d3391f3a149969aR200

model/nh_utils.F90
```diff
@@ -309,7 +308,7 @@ subroutine update_dz_d(ndif, damp, hord, is, ie, js, je, km, ng, npx, npy, area,
 
   enddo
 
-!$OMP parallel do default(none) shared(is,ie,js,je,km,ws,zs,zh,rdt)
+!$OMP parallel do default(none) shared(is,ie,js,je,km,ws,zs,zh,rdt,dz_min)
   do j=js, je
      do k=2, km+1
```
junwang-noaa, line 313: Same here

bensonr:
@XiaqiongZhou-NOAA - I didn't try using the existing PR because I was
concerned the different commit history might throw the synchronization
off.  Plus, I want the dz_min commit to come in by itself as there is a
possibility of it coming back into master.

model/nh_utils.F90
bensonr: After discussions with @junwang-noaa, I have included the master version of the height monotonicity constraint for update_dz_*

lharris4:
Hi, all. I believe that Xi and I had been discussing this earlier: for
height processes, you always want to go bottom-up.

Thanks,
Lucas

junwang-noaa:
Yes, that (go bottom-up) is in the latest dev/emc branch at EMC fork, the
same as GFDL master branch

bensonr:
I remember the discussion and am glad @junwang-noaa  brought it to my attention when reviewing the dz_min changes - even though it wasn't part of this particular PR.

model/nh_utils.F90
bensonr: @junwang-noaa can you take one more look and approve if this is now good - thanks!

junwang-noaa (approved):
Looks good to me, thanks.

PR #87, opened 2021-03-29
Fix compiling error for GNU

XiaqiongZhou-NOAA:
**Description**

Fix the compiling error with GNU

Fixes # (issue)

It was tested on Hera with the GNU compiler for UFS.

**Checklist:**

- [ X] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own code
- [X ] My changes generate no new warnings
- [ X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/model/fv_dynamics.F90
+++ b/model/fv_dynamics.F90
@@ -583,7 +583,11 @@ contains
   mdt = bdt / real(k_split)
 
   if ( idiag%id_mdt > 0 .and. (.not. do_adiabatic_init) ) then
+#ifdef __GFORTRAN__
+!$OMP parallel do default(none) shared(is,ie,js,je,npz)
+#else
 !$OMP parallel do default(none) shared(is,ie,js,je,npz,dtdt_m)
+#endif
        do k=1,npz
           do j=js,je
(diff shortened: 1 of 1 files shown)


bensonr (approved):

laurenchilutti (approved):

climbfuji (commented):
I talked to the Intel developers just last week about this problem. Recall that the Intel compiler is wrong (as their developers confirmed) and that GNU is doing the right thing. Intel said that it should be fixed in one of the next versions of their Fortran compiler, but I have no information yet which version that will be.

bensonr:
@climbfuji - thanks for the information.  I'm concerned this will show up only in the OneAPI ifx stream and not be back-ported to the soon to be retired ifort.

climbfuji:
> @climbfuji - thanks for the information. I'm concerned this will show up only in the OneAPI ifx stream and not be back-ported to the soon to be retired ifort.

Yes, this was my concern as well. Will report back when I hear from Intel.

junwang-noaa (approved):

PR #103, opened 2021-05-24
CPP/syntax errors found when running FV3 under NCAR CESM

jtruesdal:
**Description**
While porting FV3 to the NCAR CESM framework we ran into issues with the MULTI_GASES and GLOBAL_CFL CPP definitions.  Since we are not using GFS physics and have implemented some FV3 features like MULTI_GASES in our own way, we are most likely hitting bugs while traversing code paths that are not typically run. These changes were tested in our port of FV3 but not in any publically released version by GFDL. These changes should be bit for bit when tested in your normal configuration but we were unable to verify that and they should be run through regression testing.  I'm happy to work with your SE's to come up with a testing strategy that will make the PR process easier.  NCAR CESM is currently running with a fork of FV3 which contains these bug fixes and a few other code modifications.  It is our hope to work with GFDL to help generalize the portability of FV3 where we can use the official GFDL repository directly as our FV3 external. 

Files Changed:

. model/fv_dynamics.F90 
--- Code didn't compile/run when MULTI_GASES was not defined.  We fixed the issue by additional CPP definitions of MULTI_GASES as well as modifications to fv_tracer2d.F90

. model/fv_sg.F90
---  Missing enddo statement

. model/fv_tracer2d.F90
---MULTI_GASES CPP updates
--- cmax not set correctly when GLOBAL_CFL is not defined.  **Here is a fix that we are running with but will need guidance for an official fix.**

. tools/external_ic.F90
--- use of GFS physics code should be wrapped by if defined (GFS_PHYS)

. tools/fv_diag_column.F90
. tools/fv_diagnostics.F90
--- I believe nlunit was undefined in these routines when running without INTERNAL_FILE_NML defined.

Include a summary of the change and which issue is fixed. Please also include
relevant motivation and context. List any dependencies that are required for
this change.

Fixes # (issue) - NA

**How Has This Been Tested?**
This code is being run in the NCAR Community Atmosphere Model and is compiled with Intel.  If there is a public model that is supported on our systems (Cheyenne or Linux cluster) I could test these further or perhaps we can coordinate with DTC to help validate the changes if needed. Some of these issues are compile-time errors and others where caught at runtime when trying to use an uninitialized variable.

We are running the hydrostatic option for FV3 and these are the relevant CPP variables that were undefined.
GFS_PHYS not defined
MULTI_GASES not defined
GLOBAL_CFL not defined
INTERNAL_FILE_NML not defined

**Checklist:**

I'm unfamiliar with the style guidelines for coding and will need a pointer, I just tried to follow what was already there.

- [    ] My code follows the style guidelines of this project
- [ X ] I have performed a self-review of my own code
- [    ] I have commented my code, particularly in hard-to-understand areas
- [    ] I have made corresponding changes to the documentation
- [    ] My changes generate no new warnings
- [    ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/model/fv_dynamics.F90
+++ b/model/fv_dynamics.F90
@@ -274,7 +274,11 @@ contains
       integer :: rainwat = -999, snowwat = -999, graupel = -999, cld_amt = -999
       integer :: theta_d = -999
       logical used, do_omega
+#ifdef MULTI_GASES
       integer, parameter :: max_packs=13
+#else
+      integer, parameter :: max_packs=12
+#endif
       type(group_halo_update_type), save :: i_pack(max_packs)
       integer :: is,  ie,  js,  je
       integer :: isd, ied, jsd, jed
@@ -630,7 +634,9 @@ contains
          enddo
       enddo
       if ( flagstruct%trdm2 > 1.e-4 ) then
+#ifdef MULTI_GASES
          call start_group_halo_update(i_pack(13), dp1, domain)
+#endif
       endif
 
       if ( n_map==k_split ) last_step = .true.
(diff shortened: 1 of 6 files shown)


lharris4:
Hi, John. Usually FV3 is compiled with GLOBAL_CFL and MULTI_GASES turned
off so it should run fine without these options. MULTI_GASES was
implemented at EMC so they can answer questions about that (this doesn't
exist in our fork).

INTERNAL_FILE_NML is to use the FMS functionality for namelists. You may
not need this in CESM if you have your own methods.

Thanks,
Lucas

junwang-noaa:
John,

I could not find the "MULTI_GASES" in the fv_tracer2d.F90,

https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/dev/emc/model/fv_tracer2d.F90

May I ask which version you are using? Also we do have tests running with
and without " MULTI_GASES" defined, we do not see any issues.

XiaqiongZhou-NOAA:
Henry.Juang is the developer of multi_gases. It is better to ask him to review the code.

jtruesdal:
Jun, I think the original error I ran into was a runtime flag for use of an
uninitialized value related to i_pack(13) when running without MULTI_GASES
defined.  The additional MULTI_GASES cpp definition were applied to
fv_dynamics.F90 and fv_tracer2d.F90 to work around that error. We've been
running with this change for a while but I'll go back through my runs and
see if I can pin down the issue and provide the actual error.

PR #120, opened 2021-07-27
Add WAM molecular diffusion 

XiaqiongZhou-NOAA:
**Description**

Molecular diffusion developed by Henry Juang is added as a compiling option to replace Rayleigh Damping to configure FV3WAM with the model top reaching several hundreds km. 

Fixes # (issue)

**How Has This Been Tested?**
It was tested on Hera with Intel compiler and will tested on other machines following UFS regression test rules.

**Checklist:**

- [X ] I have made corresponding changes to the documentation


junwang-noaa:
@bensonr @lharris4 Would you please review this PR? Henry will retire after this week, so if you have any comments/suggestions, would you please let him/Kate know. Thanks

lharris4:
Hi, Jun. I am out of the office tomorrow but hope to look more closely
later this week. A couple of things that immediately come to mind:

- Is there any background material on what this code is supposed to
accomplish?
- Is there a reason this needs to be a compiler option, instead of a
runtime option? I could imagine this diffusion being useful for a variety
of purposes beyond deep atmospheres.
- The diff isn't very helpful for nh_core and nh_utils. What was changed in
these files?

@Henry Juang - NOAA Federal ***@***.***> Congratulations on your
upcoming retirement!

Thanks,
Lucas

model/multi_gases.F90
```diff
@@ -17,6 +17,9 @@
 !* You should have received a copy of the GNU Lesser General Public
 !* License along with the FV3 dynamical core.
 !* If not, see <http://www.gnu.org/licenses/>.
+!* --------- Default gas values for wam ----------------------------------------
+!* R_n2+= 295.3892, R_h2o= 461.50, R_o3=173.2247, R_o= 519.674, R_o2=259.8370
+!* Cpn2+=1031.1083, Cph2o=1846.00, Cpo3=820.2391, Cpo=1299.185, Cpo2=918.0969
```
bensonr, line 22: These lines need to be moved outside of the license block.

bensonr (changes requested):
I provide this review with the caveat that @lharris4 will need to weigh in as well.

There is no need to have all the #ifdefs in atmosphere.F90.  The variable "time_total" is already an optional argument to fv_dynamics and the way it is used as it traverses the call tree will not impact anything if it is "always present" in the UFS top level atmospheric dynamics driver.

If molecular diffusion is added as a runtime option, as suggested by @lharris4, many of the #ifdefs can be removed and handled with standard if-clauses.

The molecular diffusion logic in dyn_core.F90 (1256-1405) and fv_dynamics.F90 (771-829) and supporting logic should be encapsulated in subroutines which should live in the newly added molecular_diffusion.F90.  In the current setup with #ifdefs, the routines can be simple stubs in the absence of the macro.

id_ow and id_o2w are only used in fv_diagnostics.F90 and should be defined in fv_diagnostics.h and not the diag_type within fv_arrays.F90.

Please fix nh_utils.F90, sw_core.F90, tp_core.F90 and external_ic.F90 so we can see only the true diffs and not the whole file as a diff.

lharris4:
Hi, Rusty. I think you've covered most of it. It will make it much easier
to make more specific comments on the implementation once the code is
cleaned up and the diffs fixed. It would also help to have some more
information on the background motivation and the design of the algorithms
for this implementation.

Thanks,
Lucas

model/multi_gases.F90
hmhjuang: It is okay to use if-clauses though MULTI_GASES and MOLECULAR_DIFFUSION are designed specific for WAM.  For general use, some codes in molecular diffusion to compute diffusion coefficients need to add more general gases through multi_gases namelist.  Since MD and MG have lots ifdef, I suggest we leave them for next PR to change to if-clauses.

hmhjuang: Thanks to point out locations to  change, we can fix them.

hmhjuang:
Lucas, thanks for your congratulation.

model/multi_gases.F90
XiaqiongZhou-NOAA: > I provide this review with the caveat that @lharris4 will need to weigh in as well.
> 
> There is no need to have all the #ifdefs in atmosphere.F90. The variable "time_total" is already an optional argument to fv_dynamics and the way it is used as it traverses the call tree will not impact anything if it is "always present" in the UFS top level atmospheric dynamics driver.

#ifdef for time_total is removed in atmosphere.F90
> 
> If molecular diffusion is added as a runtime option, as suggested by @lharris4, many of the #ifdefs can be removed and handled with standard if-clauses.

Please suggest if it is OK to keep it is a compiling option.
> 
> The molecular diffusion logic in dyn_core.F90 (1256-1405) and fv_dynamics.F90 (771-829) and supporting logic should be encapsulated in subroutines which should live in the newly added molecular_diffusion.F90. In the current setup with #ifdefs, the routines can be simple stubs in the absence of the macro.

I guess this depends on if #ifdef is changed, right?

> 
> id_ow and id_o2w are only used in fv_diagnostics.F90 and should be defined in fv_diagnostics.h and not the diag_type within fv_arrays.F90.
> 
id_ow and id_o2w are defined in fv_diagnostics.h  as id_o3w.

> Please fix nh_utils.F90, sw_core.F90, tp_core.F90 and external_ic.F90 so we can see only the true diffs and not the whole file as a diff.
The format difference is fixed for the above files.

hmhjuang:
I have not enough time to write a detail note about molecular diffusion (MD), allow me to write a brief note and usage of its namelist variables here. So Kate may use this for add documentation in the code. Thanks, Kate!

*Brief Note*

       Molecular Diffusion (MD) implemented in this PR for FV3 was originally implemented in GSM WAM which is an operational product component of WAM-IPE Forecast System (WFS).  MD is based on physical diffusion with physical coefficients using WAM required gases, including water vapor, oxygen O2, O, O3, and N2.   MD in GSM WAM can be very effectively stable integration with strong wind speeds of 300 - 400 m/s in thermosphere without using Rayleigh Friction (RF).  In FV3, to have stable integration without MD, RF needs to be tau=0.01 which damps wind to be only 40 m/s. Due to explicit scheme, we cannot run FV3WAM with MD and without RF, however, we can use RF with tau=-50 days (negative means using GFS vertical profile for RF) to have wind speed of 400 m/s in thermosphere with MD for stable integration.
        In GSM WAM, MD was implemented in spectral space with global mean (non local) coefficient in implecity scheme. In FV3, we implemented MD as a local coefficient explicit scheme.  Since it is explicit scheme, there is a limitation of the coefficient (coefmax) to have a stable integration similar to other diffusion in FV3.  The physical coefficients are diffusivity for gas tracers, conductivity for temperature, and viscosity for 3D wind. 
        To save time, there is an namelist optional integer variable  (md_layers) to decide how many layers in the top of FV3 to apply MD, not entire model layers. Usually we have md_layers of to number of model layers to cover mesosphere to thermosphere. For current L149, we can give md_layers=70. 
        In GSM WAM, the thermosphere is very isotherm in vertical global mean, but in FV3 WAM, the initialized temperature profile is decreasing with height, to avoid this problem, the global horizontal mean temperature adjustment to isotherm in thermosphere is implemented.  The adjustment is from lower of thermosphere where temperature increasing with height then unchanging above. The starting adjustment layer is given as a namelist integer variable md_tadj_layers=25 in current L149 configuration.
       The remaining issues for MD are (1) md_tadj_layers can be very sensitive to different model physics, works with Zhao-Carr but problem with Microphysics so far (2) n_sponge has to turn off, and (3) initialize to decreasing temperature with height in thermosphere, at least, should be checked and fixed.

*Summary of namelist*
md_wait_hour : delay MD starting time, no need, can be removed.
md_layers : number of model top layers for MD, 70 is default for L149
md_tadj_layers : isonthem adjustment, 25 default for L149, 0 means off
tau_diff  : weight to diffusivity for all gas tracers, 1.0 is default, 0.0 is off
tau_cond : weight of conductivity for T, 1.0 is default, 0.0 is off
tau_visc : weight to viscosity for U, V, w, 1.0 is default, 0.0 is off

* other related for WAM* due to thermosphere..
n_sponge=-1
do_sat_adj=.f.
Both can be checked and improved later.

XiaqiongZhou-NOAA:
@lharris4 @bensonr 
Lucas, Rusty: Can you review the code again? I did revise most of what Rusty suggested except the change from the compiling option to runtime option for MOLECULAR DIFFUSION due to the time being.  Is it OK that we keep it this way and change it later?

model/multi_gases.F90
XiaqiongZhou-NOAA: > These lines need to be moved outside of the license block.

revised

model/dyn_core.F90
```diff
@@ -1234,6 +1253,156 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_map, n_split, zvir, cp,
 
     endif
 
+#ifdef MOLECULAR_DIFFUSION
```
lharris4, line 1256: I really think this all needs to be put in a separate subroutine instead of leaving it in-line in dyn_core()/

model/dyn_core.F90
```diff
@@ -1352,6 +1521,54 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_map, n_split, zvir, cp,
   enddo   ! time split loop
   first_call=.false.
 !-----------------------------------------------------
+
```
lharris4, line 1524: Could this commented code be deleted?

lharris4 (commented):
Hi, Kate. Is there an immediate need to merge the code today? I think it would be best to create the subroutines and remove as many of the #ifdefs as possible before we merge. I am concerned that anything that is checked in will be very hard to get back out.

model/dyn_core.F90
XiaqiongZhou-NOAA: > Hi, Kate. Is there an immediate need to merge the code today? I think it would be best to create the subroutines and remove as many of the #ifdefs as possible before we merge. I am concerned that anything that is checked in will be very hard to get back out.


Lucas:
I am thinking if it can be merged today, I can leave the problems to WAM new hiring :).  No problem. I can work on this if it is needed.

model/fv_dynamics.F90
```diff
@@ -1097,16 +1198,24 @@ end subroutine Rayleigh_fast
 
 
 
- subroutine Rayleigh_Super(dt, npx, npy, npz, ks, pm, phis, tau, u, v, w, pt,  &
+ subroutine Rayleigh_Super(dt, npx, npy, npz, ks, pm, phis, tau, tau_w, u, v, w, pt,  &
+#ifdef MULTI_GASES
+                           q, ncnst,  &
+#endif
```
bensonr, line 1204: Why insert this in the middle of the argument list and not at the end?

model/fv_dynamics.F90
```diff
@@ -521,9 +535,16 @@ subroutine fv_dynamics(npx, npy, npz, nq_tot,  ng, bdt, consv_te, fill,
 !            call Ray_fast(abs(dt), npx, npy, npz, pfull, flagstruct%tau, u, v, w,  &
 !                          dp_ref, ptop, hydrostatic, flagstruct%rf_cutoff, bd)
 !         else
-             call Rayleigh_Super(abs(bdt), npx, npy, npz, ks, pfull, phis, flagstruct%tau, u, v, w, pt,  &
+             call Rayleigh_Super(abs(bdt), npx, npy, npz, ks, pfull, phis, flagstruct%tau, flagstruct%tau_w, u, v, w, pt,  &
+#ifdef MULTI_GASES
+                  q, ncnst,  &
+#endif
```
bensonr, line 541: Why insert this in the middle of the argument in the list instead of at the end?

model/fv_dynamics.F90
```diff
@@ -1160,23 +1275,30 @@ subroutine Rayleigh_Super(dt, npx, npy, npz, ks, pm, phis, tau, u, v, w, pt,  &
...
 
           do k=1, ks+1
              if( is_master() ) write(6,*) k, 0.01*pm(k)
           enddo
-          if( is_master() ) write(6,*) 'Rayleigh friction E-folding time (days):'
+          if( is_master() ) write(6,*) 'Rayleigh_Super E-folding time (mb days):'
           do k=1, npz
              if ( pm(k) < rf_cutoff ) then
-                 if ( tau < 0 ) then ! GSM formula
+                 if ( tau < 0 ) then !< GSM formula
```
bensonr, line 1296: Please remove this trailing space and all throughout any changes you've made.

model/fv_dynamics.F90
```diff
@@ -1097,16 +1198,24 @@ end subroutine Rayleigh_fast
...
+    real, intent(in):: tau_w            !< time scale (days) for w 
     real, intent(in):: cp, rg, ptop, rf_cutoff
     real, intent(in),  dimension(npz):: pm
     integer, intent(in):: npx, npy, npz, ks
     logical, intent(in):: hydrostatic
+#ifdef MOLECULAR_DIFFUSION
+    real, intent(in):: conserve
+#else
     logical, intent(in):: conserve
+#endif
```
bensonr, line 1218: Changing the type of a variable in this manner and using it below is dangerous.  Especially since you are using a completely different variable when calling the routine.

model/fv_dynamics.F90
```diff
@@ -747,8 +768,68 @@ subroutine fv_dynamics(npx, npy, npz, nq_tot,  ng, bdt, consv_te, fill,
                      inline_mp, flagstruct%c2l_ord, bd, flagstruct%fv_debug, &
                      flagstruct%moist_phys)
 
+#ifdef MOLECULAR_DIFFUSION
+! do thermosphere adjustment if it is turned on and at last_step.
```
bensonr, line 772: Same as @lharris4 comment above regarding turning this block of code into a subroutine within the newly created molecular diffusion source file and not inlined here.

model/dyn_core.F90
XiaqiongZhou-NOAA: They are moved in a subroutine now.

model/fv_dynamics.F90
XiaqiongZhou-NOAA: I did not change this as many "multi_gases" are doing this way.

XiaqiongZhou-NOAA:
@junwang-noaa @hmhjuang @lharris4 @bensonr 
I think I revised all based on Lucas and Rusty's comments and suggestions. Please review again. 

@hmhjuang:
Please check if the revision is right. I have run the test. It can reproduce the results before my revision.

bensonr:
@XiaqiongZhou-NOAA - I will review it soon.  BTW, there is pressure for us to get the regional nesting work into dev/emc ASAP, therefore this will have to wait until that is done.  It planned to be ready by Friday 13 August to coincide with necessary updates in FMS 2021.03 being released that day.

hmhjuang (approved):

XiaqiongZhou-NOAA:
@junwang-noaa @lharris4 @bensonr @hmhjuang 
Lucas, Rusty: I have merged latest GFDL dev/emc to my branch and tested. Could you review the code again?

bensonr (commented):
Also, there are a few source files where you are adding in whitespace at the end of a line.  Could you please make sure to remove all unnecessary whitespace addition.  In fv_control.F90 for example, I now see a difference on the lines defining consv_te and tau which is related to extra whitespace at the end of the line.

tools/fv_diagnostics.F90
```diff
@@ -1119,6 +1119,14 @@ subroutine fv_diag_init(Atm, axes, Time, npx, npy, npz, p_ref)
             'vertical ice water flux', 'kg/m**2/s', missing_value=missing_value )
        id_o3w = register_diag_field ( trim(field), 'o3w', axes(1:3), Time, &
             'vertical ozone flux', 'kg/m**2/s', missing_value=missing_value )
+#ifdef MULTI_GASES
+       id_ow = register_diag_field ( trim(field), 'spow', axes(1:3), Time, &
+            'vertical ozone flux', 'kg/m**2/s', missing_value=missing_value )
+       id_o2w = register_diag_field ( trim(field), 'spo2w', axes(1:3), Time, &
+            'vertical ozone flux', 'kg/m**2/s', missing_value=missing_value )
+       id_o3w = register_diag_field ( trim(field), 'spo3w', axes(1:3), Time, &
+            'vertical ozone flux', 'kg/m**2/s', missing_value=missing_value )
```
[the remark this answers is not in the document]
XiaqiongZhou-NOAA: > Also, there are a few source files where you are adding in whitespace at the end of a line. Could you please make sure to remove all unnecessary whitespace addition. In fv_control.F90 for example, I now see a difference on the lines defining consv_te and tau which is related to extra whitespace at the end of the line.


extra whitespaces are removed.

model/fv_dynamics.F90
XiaqiongZhou-NOAA: This is not changed. There are many similar 

> Also, there are a few source files where you are adding in whitespace at the end of a line. Could you please make sure to remove all unnecessary whitespace addition. In fv_control.F90 for example, I now see a difference on the lines defining consv_te and tau which is related to extra whitespace at the end of the line.



> You are computing and passing time_total to fv_dynamics. It would be easier to use the fv_time variable from fv_diagnostics_mod and the get_time function to compute the time_total value where needed. There is a plan to utilize the existing Atm%Time variable for this purpose in the future.

time_total is calculated in fv_dynamics now.

XiaqiongZhou-NOAA:
@bensonr 

> Also, there are a few source files where you are adding in whitespace at the end of a line. Could you please make sure to remove all unnecessary whitespace addition. In fv_control.F90 for example, I now see a difference on the lines defining consv_te and tau which is related to extra whitespace at the end of the line.

extra whitespaces are removed

XiaqiongZhou-NOAA:
> You are computing and passing time_total to fv_dynamics. It would be easier to use the fv_time variable from fv_diagnostics_mod and the get_time function to compute the time_total value where needed. There is a plan to utilize the existing Atm%Time variable for this purpose in the future.

time_total is calculated in fv_dynamics now.

model/fv_dynamics.F90
XiaqiongZhou-NOAA: This part is not changed since similar #ifdef was used in many other places. I guess it is to avoid the #else part.

model/fv_arrays.F90
```diff
@@ -671,6 +682,7 @@ module fv_arrays_mod
                               !< original value before entering the physics; a value of 0.7 roughly
                               !< causes the energy fixer to compensate for the amount of energy changed
                               !< by the physics in GFDL HiRAM or AM3.
+   real    :: tau_w = 0.      !< Time scale (in days) for Rayleigh friction applied to vertical wind
```
bensonr, line 685: I believe the description for tau_w and tau need to be updated.  If I'm reading the code correctly, when tau_w is defined as zero (the default), tau still controls the RF damping timescale for the vertical winds.  But if tau_w is anything other than zero, it is now and independent timescale for the RF damping of the vertical winds and tau no longer has any effect.  Also think there needs to be some discussion of what happens in a hydrostatic system.  Is this something that should only be used in WAM simulations or does it have general application?

bensonr (changes requested):

model/fv_arrays.F90
XiaqiongZhou-NOAA: @bensonr @hmhjuang 
You are right. 
It should work for any general application. I believe it does not affect hydrostatic system. I will add more description.

lharris4:
description for tau_w looks OK

bensonr:
@XiaqiongZhou-NOAA - I believe we are almost ready for this one.  It looks like there is a conflict with the makefile, which I think has been deleted.

XiaqiongZhou-NOAA:
I have not merged the latest GFDL dev/emc to my branch. I will do it soon.

XiaqiongZhou-NOAA:
@bensonr The PRs in fv3atm and ufs are planned to merge Tuesday (10/12)

XiaqiongZhou-NOAA:
@bensonr @lharris4
Lucas is right. 
Also I am afraid I have to change the way to assign values to these arrays. There is a memory issue when running compiling jobs on DELL. This revision can fix it (even though I do not understand why).

XiaqiongZhou-NOAA:
@bensonr  This is not a bug for gnu/9.2.0 on Hera. All RTs  can be compiled OK there.

XiaqiongZhou-NOAA:
I will revise the range definition back and keep the original form. Another method will be used to fix the compiling issue on DELL: compiling fv_diagnostics.F90 with -O0 instead of -O2.

PR #126, opened 2021-08-04
Ensure tracer mass conservation during preconditioning

### LINKED ISSUE #125: Total tracer mass is not conserved after adiabatic preconditioning
**Describe the bug**
Adiabatic preconditioning of initial conditions (`na_init>0`) does not conserve the total mass of prognostic tracers.
This was discovered while performing experiments that involved cycling aerosol tracers using new atmospheric initial conditions at each cycle.

This issue is affecting P7-targeted prognostic aerosol experiments.

**To Reproduce**
Set `na_init` to a positive integer (usually `1`), then compute the total tracer mass before and after preconditioning and take the difference.

**Expected behavior**
I suggest rescaling mixing ratios of prognostic tracers every time `Atm%delp` is adjusted to account for the change in the layer's total mass.

rmontuoro:
**Description**

This PR introduces changes aimed at ensuring mass conservation for all prognostic tracers during adiabatic preconditioning.

Mixing ratios are rescaled by `Atm%delp(1)/Atm%delp(2)`, where `(1)` and `(2)` indicate the original and updated `delp` array after nudging. This correction is applied after each `delp` update to all tracers, except for `sphum` if `nudge_qv` is `.true.`.

Fixes #125 

**How Has This Been Tested?**

The changes were tested on Hera/Intel, running both standalone FV3 and coupled FV3-GOCART.

**Checklist:**


bensonr (commented):
it looks like this update will change all baselines when the adiabatic pre-conditioning is enabled.  Is that the intent?

rmontuoro:
@bensonr - I agree, this update has the potential to alter all baselines created with adiabatic preconditioning due to small changes in the MP tracers. It does change the baselines of tests including prognostic aerosols.

The intent is to enforce mass conservation for all prognostic tracers. Whether this should be done or not for hydrometeors from external ICs is debatable, since we're altering the input atmospheric state anyway during preconditioning.

lharris4:
Hi, Raffale. If this is used to modify only tracers n > nwat then it will
not change the global dry mass. Since you only need to reset  the mass of
non-hydrometeor constituents, this will achieve your result without
changing the dry mass.

The variable delp is the total mass of the grid cell, dry air + water
vapor + the nwat microphysical constituents.

Lucas

rmontuoro:
Thank you, @lharris4. Indeed we could only adjust the mixing ratios of tracers > `nwat` and neglect the hydrometeors.

Would this be your recommendation? Should we attempt to conserve the hydrometeor total mass as ingested from ICs or is it safe to neglect these adjustments, since the atmospheric state is being nudged (especially if `nudge_qv` is `.true.`)?

rmontuoro:
Thanks @lharris4, this makes sense to me. I've just pushed a change to skip the hydrometeors when enforcing mass conservation.

driver/fvGFS/atmosphere.F90
```diff
@@ -2010,6 +2016,28 @@ subroutine adiabatic_init(zvir,nudge_dz,time)
...
+     if ( nq > Atm(mygrid)%flagstruct%nwat ) then
+       ! Adjust prognostic tracer mixing ratios for consistency with updated layer mass
+!$omp parallel do default (none) &
+!$omp              shared (jsc, jec, isc, iec, npz, nq, Atm, dp0, mygrid) &
+!$omp             private (i, j, k, n)
+       do n = Atm(mygrid)%flagstruct%nwat + 1, nq
+         do k = 1, npz
+           do j = jsc, jec
+             do i = isc, iec
+               Atm(mygrid)%q(i,j,k,n) = Atm(mygrid)%q(i,j,k,n) * (1. + wt * (1. - dp0(i,j,k) / Atm(mygrid)%delp(i,j,k)))
```
bensonr, line 2032: Besides condensates, are there any other tracer types that shouldn't be scaled in this fashion - e.g. number concentrations?

rmontuoro: Any tracer type whose concentration is expressed as ratio of a given quantity to air mass will need to be adjusted.

rmontuoro:
This PR will affect the predicted concentrations of non-hydrometeor prognostic tracers if they are included in the ICs.

climbfuji:
> This PR will affect the predicted concentrations of non-hydrometeor prognostic tracers if they are included in the ICs.

We should check then how the Thompson aerosol number concentrations are affected. They are interpolated from the climatology (or from real observational data) and added to the ICs by `chgres_cube` for Thompson MP.

rmontuoro:
@climbfuji - Thompson's aerosol number concentrations will be rescaled, if included in the ICs, since they are located after `nwat`.

climbfuji:
> @climbfuji - Thompson's aerosol number concentrations will be rescaled, if included in the ICs, since they are located after `nwat`.

That's what I thought, we should check how much they change though. It doesn't matter for the crude aerosol climatology data that is used most of the time, but for real-time fire emission data we should check (Georg's group is experimenting with this). I don't think it will be a problem, but a sanity check within the next few weeks would be good (we don't need to delay the commit).

lharris4:
Is there a simple way to distinguish between tracers and number
concentrations? We could (for example) look at the name of the tracer
defined in field_table.

climbfuji:
> Is there a simple way to distinguish between tracers and number concentrations? We could (for example) look at the name of the tracer defined in field_table.
> […](#)

I am not sure if we should base decisions on tracer names, rather on attributes that specify the nature of a field listed, or what to do with the field. Such a capability would also allow us to say that `cld_amt` shouldn't be advected instead of having to set `dnats=1` and relying on it being the last item in `field_table`.

lharris4:
I am hoping to make better use of the tracer_manager features in
field_table; for example, we could add an attribute that defines how a
variable should be transformed. I think we can just simply add the
attribute to the field_table without needing to make any changes in FMS; we
would then just check that attribute within the model.

I would also hope that we can get rid of "dnats" and just use the ability
to set a field to be diagnostic (non-advected), also in field_table.

climbfuji:
> I am hoping to make better use of the tracer_manager features in field_table; for example, we could add an attribute that defines how a variable should be transformed. I think we can just simply add the attribute to the field_table without needing to make any changes in FMS; we would then just check that attribute within the model. I would also hope that we can get rid of "dnats" and just use the ability to set a field to be diagnostic (non-advected), also in field_table.
> […](#)

That sounds great to me, Lucas. This would also fit nicely with ongoing CCPP work. Please let me know if there is something we can contribute. Thanks!

lharris4:
Hi, Dom. I think this is something that Raffale could pretty easily try
himself. He should be able to add a new tag in each field_table to indicate
whether something is a hydrometeor, constituent, or

An example for how to use field_table this way. You can define arbitrary
pairs of strings in field_table to describe quantities.
Here is an example entry in field_table for AM4:

"tracer","atmos_mod","co"
"longname","CO"
"units","VMR"
"emissions", "file"
"emissions3d", "file"
"dry_deposition", "file"
"init_conc", "file"
"convection","all"
/

Inside the code, you can use the routine tracer_manager_mod::query_method
to get the value for a given key, and whether it exists. See an example
here:

https://github.com/NOAA-GFDL/atmos_phys/blob/46817a12232ed54d1a5d5a4394a695621f575812/atmos_shared/tracer_driver/atmos_dust.F90#L472

(The tracer_driver within AM4 has a number of other uses; and this routine
is used extensively in MOM6, but isn't yet used within FV3.)

Please let us know if there are any issues you might see.

Thanks
Lucas

rmontuoro:
@lharris4, @climbfuji - The suggested approach was implemented for aerosol tracers in fv3atm through PR [#310](https://github.com/NOAA-EMC/fv3atm/pull/310#issue-651065377). This PR introduced the additional `tracer_usage` keyword to recognize chemistry tracers and their type (prognostic or diagnostic):
```
      "tracer_usage", "chemistry"                    (default, prognostic)
      "tracer_usage", "chemistry", "type=diagnostic" (diagnostic)
 ```

I suggest having a discussion in a separate GitHub issue to coordinate and standardize keyword names and usage in this and other related code bases.

I agree with @climbfuji that getting rid of `dnats` would be ideal, however it will involve code refactoring for the dycore, possibly including internal tracer remapping.

bensonr:
On Mon, Aug 23, 2021 at 7:14 PM Raffaele Montuoro ***@***.***>
wrote:

> @lharris4 <https://github.com/lharris4>, @climbfuji
> <https://github.com/climbfuji> - The suggested approach was implemented
> for aerosol tracers in fv3atm through PR #310
> <https://github.com/NOAA-EMC/fv3atm/pull/310#issue-651065377>. This PR
> introduced the additional tracer_usage keyword to recognize chemistry
> tracers and their type (prognostic or diagnostic):
>
>       "tracer_usage", "chemistry"                    (default, prognostic)
>       "tracer_usage", "chemistry", "type=diagnostic" (diagnostic)
>
> I suggest having a discussion in a separate GitHub issue to coordinate and
> standardize keyword names and usage in this and other related code bases.
>
> I agree with @climbfuji <https://github.com/climbfuji> that getting rid
> of dnats would be ideal, however it will involve code refactoring for the
> dycore, possibly including internal tracer remapping.
>

The dycore already has a mode where dnats is obsolete.  You may have
noticed the Atm%qdiag array.  This is used within the GFDL climate models
to hold diagnostic non-advected tracers.  The FMS tracer manager already
segments the tracers by type with advected numbered sequentially and
diagnostic/non-advected beginning after the advected ones  The order of
diagnostic and prognostic within the field_table does not matter, iirc.

lharris4:
Hi, all. That is great. I have been hoping to be rid of dnats for a long
time. We might do this in the next FV3 release.

Lucas

bensonr (approved):

yangfanglin (approved):
looks fine to me

driver/fvGFS/atmosphere.F90
junwang-noaa: @rmontuoro I am wondering what happens to the prognostic subgrid scale turbulent kinetic energy "sgs_tke" which is also after nwat, are you also going to adjust it? Have you done testing to see if the number concentration fields or the tke field should be adjusted? Can you show some of your results and test configuration from this update? I am not sure if we have restriction that all the prognostic tracers after nwat are expressed as ratio to air mass.

bensonr:
@junwang-noaa can you review and give your approval/feedback - thanks

climbfuji:
> On Mon, Aug 23, 2021 at 7:14 PM Raffaele Montuoro ***@***.***> wrote: @lharris4 <https://github.com/lharris4>, @climbfuji <https://github.com/climbfuji> - The suggested approach was implemented for aerosol tracers in fv3atm through PR #310 <[NOAA-EMC/fv3atm#310 (comment)](https://github.com/NOAA-EMC/fv3atm/pull/310#issue-651065377)>. This PR introduced the additional tracer_usage keyword to recognize chemistry tracers and their type (prognostic or diagnostic): "tracer_usage", "chemistry" (default, prognostic) "tracer_usage", "chemistry", "type=diagnostic" (diagnostic) I suggest having a discussion in a separate GitHub issue to coordinate and standardize keyword names and usage in this and other related code bases. I agree with @climbfuji <https://github.com/climbfuji> that getting rid of dnats would be ideal, however it will involve code refactoring for the dycore, possibly including internal tracer remapping.
> The dycore already has a mode where dnats is obsolete. You may have noticed the Atm%qdiag array. This is used within the GFDL climate models to hold diagnostic non-advected tracers. The FMS tracer manager already segments the tracers by type with advected numbered sequentially and diagnostic/non-advected beginning after the advected ones The order of diagnostic and prognostic within the field_table does not matter, iirc.

@bensonr What syntax do I need to use in `field_table` for tracers that are purely diagnostic and should not be advected? 

@rmontuoro @junwang-noaa It is not clear to me whether the problem of `sgs_tke` being rescaled or not is being solved. If it is, what changes are required in `field_table` for `sgs_tke`?

And, since we don't use the `qdiag` array yet, but instead `dnats=1` for `cld_amt`, am I correct that `cld_amt` will not be rescaled?

bensonr:
@DomHeinzeller @climbfuhi - you add an attribute "tracer_type" to denote the type as diagnostic.  Here's an example:

"tracer","atmos_mod","ch3o2"
"longname","CH3O2"
"units","VMR"
"tracer_type","diagnostic"
"profile_type","fixed","surface_value=1.e-20"
/

For FV3, this will store this tracer in the qdiag array which will be included in the Statein%qgrs.  Note that these tracers are not mass adjusted going to/from the dycore.  There may also be an issue with external_ic as I don't believe it assumes any of the tracers ingested are diagnostic and there most likely will need to be additional logic to properly place these tracers into the Atm%qdiag array.

rmontuoro:
> @bensonr I had some questions on this update and did not not get reply yet. Please see below:
> 
> @rmontuoro I am wondering what happens to the prognostic subgrid scale turbulent kinetic energy "sgs_tke" which is also after nwat, are you also going to adjust it? Have you done testing to see if the number concentration fields or the tke field should be adjusted? Can you show some of your results and test configuration from this update? I am not sure if we have restriction that all the prognostic tracers after nwat are expressed as ratio to air mass.
> 
> @rmontuoro, would you please show some results about the impact to this code change on GFSv16 or coupled model and confirm the code is working as we expected.

@junwang-noaa, @bensonr - below are my answers:

- all tracers after `nwat` *will be adjusted*. Number concentrations need rescaling, too, since they are expressed as #/mass.
- the prognostic subgrid scale turbulent kinetic energy (TKE) tracer (`sgs_tke`) will also be rescaled if included in the IC file. However, my understanding is that interpolating TKE from IC is inaccurate and should be avoided. This seems to be confirmed by the following statements initializing the TKE tracer to either `0` or to a small value:
    - NGGPS ICs: https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/ab026b79497ee00530ff3eb6b59dd9b9346e8e7d/tools/external_ic.F90#L839-L849
    - ECMWF ICs: https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/ab026b79497ee00530ff3eb6b59dd9b9346e8e7d/tools/external_ic.F90#L2591-L2600
    - HRRR ICs: https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/ab026b79497ee00530ff3eb6b59dd9b9346e8e7d/tools/external_ic.F90#L1319-L1329
Note that tracer initialization is performed before preconditioning. We could re-initialize non-zero TKE tracers as above after adiabatic preconditioning if needed.
- We have been using this update in all of P7-related validation experiments for prognostic aerosols, and we would be happy to share our results once completed
- Tracer units are not enforced, but it is understood that the dycore requires mass ratios

I would be happy to run R/T again with the updated UFS weather model code to verify everything is working as expected once we decide how to move forward.

rmontuoro:
Regression testing with more recent versions of FV3 showed few abnormally high differences in U, V, and T after a 24-hour forecast run, which are currently been investigated. While such differences may not be directly caused by the small mixing ratio adjustments introduced here and may be indicative of deeper issues, I recommend holding this PR until the source of such large differences has been identified.

laurenchilutti:
@rmontuoro Should we close this PR and reopen when you are ready to revisit?

rmontuoro:
@laurenchilutti - Thanks for checking. I would like to keep this PR open, since this issue needs to be addressed in order to perform cycled runs with prognostic aerosols.

junwang-noaa:
@rmontuoro Can you explain why we still need this change for cycled runs? In the cycled run, the aerosols are coming from model restart files just like other atmosphere tracers, why do we need this adjustment? I thought fv3atm has good mass conversations without this adjustment as shown in many presentations. 

My understanding is that this is required for cold start runs with unbalanced ICs. Also I'd suggest adding the code changes as an option for aerosol tracer adjustment if that is the concern, so you can do more testing with the option turned on.

bensonr:
Closing as there has been no response for over a year.

PR #146, opened 2021-09-22
Bug fix fv_regional_bc.F90: initialize cloud amount boundary conditions only if cloud amount is a valid tracer

DomHeinzeller:
**Description**

As the title says. In `model/fv_regional_bc.F90`, initialize cloud amount boundary conditions to zero only if cloud amount is a valid tracer (i.e. if the index in the tracer array is > 0).

Fixes https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/145

Tested in the latest ufs-weather-model for regression tests that are run only sporadically (`rt_ccpp_dev.conf`), i.e. not as part of the standard Intel/GNU tests. This solves a crash with old input data that was generated for Thompson MP and doesn't have cloud amount.

**Checklist:**

Please check all whether they apply or not
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [n/a] I have commented my code, particularly in hard-to-understand areas
- [n/a] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [n/a] Any dependent changes have been merged and published in downstream modules

PR #147, opened 2021-09-23
Initializing tracers to fix the FP issue for some tracers in debug mo…

JosephMouallem:
Initializing tracers to fix the FP issue for some tracers in debug mode triggered in remap_scalar

Fixes # 
Fixes a crash triggered in hafs debug mode on some nests.

**How Has This Been Tested?**

Tested on Orion under HAFS

**Checklist:**


bensonr (approved):

laurenchilutti (approved):

junwang-noaa (approved):
Just want to mention that dev/emc was updated, maybe we need to sync this PR before merging.

bensonr:
@junwang-noaa - while it is desirable to sync prior to a merge, it isn't a requirement when they aren't touching the same files.

laurenchilutti:
@DusanJovic-NOAA @junwang-noaa Could you please review this again as changes were made following your original approval?

DusanJovic-NOAA (approved):

junwang-noaa (approved):

laurenchilutti (approved):

bensonr (approved):

climbfuji:
@bensonr @JosephMouallem @junwang-noaa @DusanJovic-NOAA @laurenchilutti @XiaqiongZhou-NOAA I think this PR wrecks havoc when Thompson microphysics is used. I noticed that after updating the GFDL_atmos_cubed_sphere submodule pointer in my PR https://github.com/ufs-community/ufs-weather-model/pull/850 to include #146, #147, #148, all regression tests using Thompson MP have different answers. The reason is that after the adiabatic init, i.e. before the first time physics get called, the number concentrations `ice_nc` and `rain_nc` the fourth-last and third-last indices in the tracer arrays have nonsense values.

#148 and #146 have nothing to do with these tracers (or their associattedd mass mixing ratios) and are not responsible for the problem.

I believe that the problem is that this PR #147 sets those tracers (and especially missing ice and rain mass mixing ratios) to `-999` instead of zero. This may be correct to indicate that they are missing, but the original initialization in `GFS_typedefs.F90` is to set them to zero, and the CCPP physics logic relies on them being zero and not -999 (not only for those two tracers, but many variables - this is not something CCPP invented, in general it used to be the case beforehand already).

Thompson MP at initialization computes number concentrations from mass concentrations, and the logic doesn't know about those -999s. See attached screenshot.

Screen Shot 2021-10-10 at 8 28 02 PM

So, the submodule pointer update to include #147 (and by this also #148) need to be put on hold until the necessary changes are made to the logic in either fv3atm or ccpp-physics (it is not clear to me on this late Sunday evening during my holidays what the best place is).

PR #149, opened 2021-10-04
Fix undeclared variable/routines in namelist IO

christopherwharrop-noaa:
**Description**

This fixes a bug in `driver/fvGFS/atmosphere.F90` that causes compilation to fail when the code is built without `-DINTERNAL_FILE_NML`.  Specifically, where the namelist is read, the variable, `unit`, was not declared and the routines `open_namelist_file` and `close_file` were not available.  This fixes those issues.

**How Has This Been Tested?**

The code was successfully compiled and run with the Intel OneAPI 2021 compiler in an Ubuntu 20.04 container.

**Checklist:**

### DIFF
--- a/driver/fvGFS/atmosphere.F90
+++ b/driver/fvGFS/atmosphere.F90
@@ -154,7 +154,8 @@ use fms_mod,                only: error_mesg, FATAL,                 &
                                   write_version_number,              &
                                   mpp_clock_id, mpp_clock_begin,     &
                                   mpp_clock_end, CLOCK_SUBCOMPONENT, &
-                                  clock_flag_default
+                                  clock_flag_default,                &
+                                  open_namelist_file, close_file
 use fms2_io_mod,            only: file_exists
 use mpp_mod,                only: mpp_error, stdout, FATAL, WARNING, NOTE, &
                                   input_nml_file, mpp_root_pe,    &
@@ -476,13 +477,13 @@ contains
    read(input_nml_file, nml=atmos_model_nml, iostat=io)
    ierr = check_nml_error(io, 'atmos_model_nml')
 #else
-   unit = open_namelist_file ( )
+   nlunit = open_namelist_file ( )
    ierr=1
    do while (ierr /= 0)
-      read  (unit, nml=atmos_model_nml, iostat=io, end=10)
+      read  (nlunit, nml=atmos_model_nml, iostat=io, end=10)
       ierr = check_nml_error(io,'atmos_model_nml')
    enddo
-10 call close_file (unit)
+10 call close_file (nlunit)
 #endif
    !write(0,'(a)') "It's me, and my physics suite is '" // trim(ccpp_suite) // "'"
    ! *DH 20210326


bensonr:
@christopherwharrop-noaa - thanks for pointing this case of incomplete cleanup of the source file.  The intended fix was to remove the option to open/read/close the nameilst and leave only the block with the internal file read.  If you want to perform that work, we'd appreciate it.  If not, we can close this PR and use the existing issue against a new PR.

christopherwharrop-noaa:
@bensonr - I don't feel comfortable doing that because I am not privy to all of the potential consequences of that to all of the UFS apps (and their many branches) downstream.  That seems like something that should be coordinated with folks more familiar with the details of the current status of app development.  I do know that the `ufs-srweather-app` branches that I am working with need that block to read the namelist. As I am not involved with the core development efforts, modifying the ufs-srweather-app to remove that usage is outside my wheelhouse, so I'd want to ask others more familiar with the development (not sure the best person) to help with that.

christopherwharrop-noaa:
I suppose any apps that are using the open/read/close the namelist will have run into this....So, they'd have to fix it.  I'm just not super comfortable messing with this because I'm not really involved with the app development at that level. Now that I think about it more carefully, it may be that the ufs-srweather-app `develop` is already using the internal file read method, and other branches just need to merge that update in.

christopherwharrop-noaa:
@bensonr - I have just verified that the current `develop` for the community version (https://github.com/ufs-community/ufs-srweather-app) also has this issue.  This is bad.  I understand the code being fixed by this PR was not intended to exist in the first place, but both the community version and organizational level forks currently rely on it. If this simple fix cannot be applied here, there needs to be broad coordination with the UFS app developers to fix this on the UFS app side (I am not even sure what, exactly, needs to be changed on the app side to fix this).

I will file an issue in the relevant places.  In the meantime, many people will need to use forks of this repo that have this fix applied.

```
[ 87%] Building Fortran object FV3/CMakeFiles/fv3dycore.dir/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90.o
/opt/ufs-srweather-app.community/src/ufs-weather-model/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90(479): error #6404: This name does not have a type, and must have an explicit type.   [UNIT]
   unit = open_namelist_file ( )
---^
/opt/ufs-srweather-app.community/src/ufs-weather-model/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90(479): error #6404: This name does not have a type, and must have an explicit type.   [OPEN_NAMELIST_FILE]
   unit = open_namelist_file ( )
----------^
Using 8-byte addressing
Using pure routines.
Using allocatable derived type array members.
Using cray pointers.
compilation aborted for /opt/ufs-srweather-app.community/src/ufs-weather-model/FV3/atmos_cubed_sphere/driver/fvGFS/atmosphere.F90 (code 1)
```

bensonr:
@christopherwharrop-noaa - this is because the -DINTERNAL_FILE_NML macro was removed from the compile system and this #ifdef was left in place as I mentioned previously.  If you remove lines 475 & 476-486 you won't have any issues with the resulting code/run.

christopherwharrop-noaa:
@bensonr - I am sorry.  I completely misunderstood what was going on here and what you were saying.  I had thought the apps needed the `#else` block of that `#ifdef`, and that if I removed it, they would not work.  I can certainly modify this PR to remove the `#ifdef`, leaving only lines 476 & 477.:

```
   read(input_nml_file, nml=atmos_model_nml, iostat=io)
   ierr = check_nml_error(io, 'atmos_model_nml')
```

I am happy to do that.  I'll do a run first and make sure nothing unexpected happens, and then update this PR.

Again, sorry for the confusion happening on my end.

bensonr:
@christopherwharrop-noaa - No need to apologize.  All that matters is we get to the proper result in the end.

christopherwharrop-noaa:
@bensonr - I have updated this PR to reflect your wishes.  Please let me know if additional changes are required.

PR #167, opened 2022-01-04
fix for 4diau with iau_filter_increments=T

jswhit:
**Description**

This PR fixes a bug in the 4DIAU code when the parameter iau_filter_increments=T.  The time bounds for the IAU interval were computed incorrectly, leading to incorrect filter weights that were applied outside the designated time range.   The 4DIAU with constant weights, and 3DIAU (with a single increment file specified) were not affected by this bug.

Fixes https://github.com/ufs-community/ufs-weather-model/issues/989

**How Has This Been Tested?**

Regression tests pass on hera, cycling DA experiment with iau_filter_increments=T and hourly increment files ran on orion.

**Checklist:**

- [ x] My code follows the style guidelines of this project
- [x ] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings

### DIFF
--- a/tools/fv_iau_mod.F90
+++ b/tools/fv_iau_mod.F90
@@ -298,7 +298,7 @@ subroutine getiauforcing(IPD_Control,IAU_Data)
    implicit none
    type (IPD_control_type), intent(in) :: IPD_Control
    type(IAU_external_data_type),  intent(inout) :: IAU_Data
-   real(kind=kind_phys) t1,t2,sx,wx,wt,dtp
+   real(kind=kind_phys) t1,t2,tnext,sx,wx,wt,dtp
    integer n,i,j,k,sphum,kstep,nstep
 
    IAU_Data%in_interval=.false.
@@ -353,7 +353,7 @@ subroutine getiauforcing(IPD_Control,IAU_Data)
    endif
 
    if (nfiles > 1) then
-      t2=2
+      tnext=2
       if (IPD_Control%fhour < t1 .or. IPD_Control%fhour >= t2) then
 !         if (is_master()) print *,'no iau forcing',IPD_Control%iaufhrs(1),IPD_Control%fhour,IPD_Control%iaufhrs(nfiles)
          IAU_Data%in_interval=.false.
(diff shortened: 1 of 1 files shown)


MingjingTong-NOAA (changes requested):
I tested the changes for both 4DIAU and 3DIAU. The fix works for 4DIAU, but not for 3DIAU. When running 3DIAU, nfiles = 1 and IPD_Control%iaufhrs(1) = IPD_Control%iaufhrs(nfiles). The 'if' condition in line 318 is never satisfied (iau_state%wt is always 0), and the if condition in line 336 is always satisfied (IAU_Data%in_interval is always false).

MingjingTong-NOAA:
The new changes works for 3DIAU now, but not 4DIAU. The problem is the change in line 357. The t2 here has different meaning. It's no longer IPD_Control%iaufhrs(nfiles). If reverting the change of line 357, it works for 4DIAU. 

I have another recommendation for 3DIAU. Please consider removing the 'if' condition 'if (IPD_Control%iau_filter_increments) ' in line 348. Otherwise, if iau_filter_increments = .false., setiauforcing will not be called, the iau increment will not be normalized by rdt.

MingjingTong-NOAA (changes requested):
The new changes works for 3DIAU now, but not 4DIAU. The problem is the change in line 357. The t2 here has different meaning. It's no longer IPD_Control%iaufhrs(nfiles). If reverting the change of line 357, it works for 4DIAU.

I have another recommendation for 3DIAU. Please consider removing the 'if' condition 'if (IPD_Control%iau_filter_increments) ' in line 348. Otherwise, if iau_filter_increments = .false., setiauforcing will not be called, the iau increment will not be normalized by rdt.

MingjingTong-NOAA:
Please ignore my second recommendation. I saw that setiauforcing is called in IAU_initialize.

laurenchilutti:
@MingjingTong-NOAA do you think this PR is acceptable or do you still think there are changes that should be made?

jswhit:
@MingjingTong-NOAA thanks for reviewing this. I see the problem you found is related to the variable t2 being redefined.  Instead of your suggested fix (which would have worked fine) I created a different variable tnext and preserved the t2 variable to use in the if test.  4DIAU should work now.

MingjingTong-NOAA:
The fix looks fine to me. The pull request can be merged. Thanks!

bensonr (approved):

junwang-noaa:
@laurenchilutti @bensonr All the RT passed in ufs-weather-model [PR#990](https://github.com/ufs-community/ufs-weather-model/pull/990), would you please commit this PR? Thanks

laurenchilutti (approved):

PR #170, opened 2022-01-28
Fix for the doubly-periodic supercell thunderstorm case 17

JosephMouallem:
**Description**

Fix for the idealized test case 17 to simulate a supercell thunderstorm in a doubly periodic domain

**How Has This Been Tested?**

Tested on Orion under SHiELD with the latest FMS and physics

**Checklist:**

### DIFF
--- a/tools/test_cases.F90
+++ b/tools/test_cases.F90
@@ -4927,11 +4927,17 @@ end subroutine terminator_tracers
               do i=is,ie
                  pt(i,j,k)   = ts1(k)
                   q(i,j,k,1) = qs1(k)
-                 delz(i,j,k) = rdgas/grav*ts1(k)*(1.+zvir*qs1(k))*(peln(i,k,j)-peln(i,k+1,j))
+!                 delz(i,j,k) = rdgas/grav*ts1(k)*(1.+zvir*qs1(k))*(peln(i,k,j)-peln(i,k+1,j))
                 enddo
              enddo
           enddo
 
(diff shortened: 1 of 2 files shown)


lharris4:
Hi, Joseph. This looks good. Thanks for taking care of this.

Lucas

lharris4 (approved):
Looks good to me.

tools/test_cases.F90
```diff
@@ -5566,7 +5570,8 @@ end subroutine superK_u
 
 
  subroutine SuperCell_Sounding(km, ps, pk1, tp, qp)
- use gfdl_cloud_microphys_mod, only: wqsat_moist, qsmith_init, qs_blend
+ use gfdl_cloud_microphys_mod, only: qsmith_init, qs_blend
+ use gfdl_mp_mod, only: wqsat_moist
```
bensonr, line 5574: I think we need to be careful about which one we use.  As I understand it, one is to be compiled with -r8 (physics) and the other with -r4 (inline dycore physics) - but maybe I've got it all wrong.

lharris4:
I believe that this code compiles fine, although perhaps qsmith_init and
qs_blend from gfdl_mp_mod should be used instead.

lharris4 (approved):
This is an appropriate update for this test. I approve this MR. Thanks for taking care of this.

bensonr (approved):

JosephMouallem:
Will be part of the upcoming public release, so I will close it.

PR #180, opened 2022-03-11
logic fixes for climate nudging

uramirez8707:
**Description**

This PR:
1. Changes the arguments of `check_resolution` in read_climate_nudge_data.F90 to `fileobj` and `field_name`
2. Moves the `get_variable_num_dimensions` and `get_variable_size` calls in read_climate_nudge_data_init to `check_resolution`

Fixes # (issue)
`get_variable_size` was being called for all the variables in the file, which _may_ result in a crash because it [cannot be called for scalars](https://github.com/NOAA-GFDL/FMS/blob/7b78a73a5ba7acf5d3d932ecfe081e5040e2c778/fms2_io/netcdf_io.F90#L1610-L1618). With these changes `get_variable_size` is now only being called for the variables the code [actually needs](https://github.com/uramirez8707/GFDL_atmos_cubed_sphere/blob/f724ade2ad8ac538c5af410387ed5e90cc7e5cc1/GFDL_tools/read_climate_nudge_data.F90#L188), and if the variable is a scalar it will crash as expected. 

**How Has This Been Tested?**
AM4p2 nudged with intel18, prod-openmp

**Checklist:**

### DIFF
--- a/GFDL_tools/read_climate_nudge_data.F90
+++ b/GFDL_tools/read_climate_nudge_data.F90
@@ -104,7 +104,7 @@ integer, intent(out) :: nlon, nlat, nlev, ntime
 !   ntime       number of time levels
 
   integer :: iunit, ierr, io
-  integer :: istat, i, j, k, n, nd, siz(4), i1, i2
+  integer :: istat, i, j, k, n, i1, i2
   character(len=32), allocatable :: fields(:)
   type(FmsNetcdfFile_t) :: fileobj
   integer, allocatable, dimension(:) :: pes !< Array of ther pes in the current pelist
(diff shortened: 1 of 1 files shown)


GFDL_tools/read_climate_nudge_data.F90
```diff
@@ -579,23 +577,35 @@ end subroutine check_axis_size
...
       call error_mesg ('read_climate_nudge_data_mod', 'incorrect array dimension one', FATAL)
    endif
-   if (axis_len(2) .ne. global_axis_size(INDEX_LAT)) then
+   if (siz(2) .ne. global_axis_size(INDEX_LAT)) then
       call error_mesg ('read_climate_nudge_data_mod', 'incorrect array dimension two', FATAL)
    endif
-   if (size(axis_len(:)) .gt. 3) then
-      if (axis_len(3) .ne. global_axis_size(INDEX_LEV)) then
+   if (nd .gt. 3) then
+      if (siz(3) .ne. global_axis_size(INDEX_LEV)) then
```
bensonr, line 604: This assumes any index beyond 3 will not need to be checked.  Should the index checking be put into a loop with limits of 1,min(nd,3) or should it go until it reaches nd or finds an unlimited?

laurenchilutti:
@uramirez8707 Could you take a look at Rusty's comment?  We want to try to get this merged in.

GFDL_tools/read_climate_nudge_data.F90
uramirez8707: @bensonr It looks like global_axis_size has a size of 3:
https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/f724ade2ad8ac538c5af410387ed5e90cc7e5cc1/GFDL_tools/read_climate_nudge_data.F90#L80

So even if it the index is greater than 3, there wouldn't be anything to compare the dimensions to.

bensonr (approved):

laurenchilutti (approved):

PR #183, opened 2022-04-05
Update CMakeLists.txt

natalie-perlin:
**Description**

To allow running on Mac OSX (Apple) machines, QUAD_PRECISION needs to be turned off

Fixes # (issue)

Replace line 35 where the option ENABLE_QUAD_PRECISION is set to ON, by the conditional statement that sets it to OFF on APPLE machines:

if(APPLE)
     option(ENABLE_QUAD_PRECISION "Enable compiler definition -DENABLE_QUAD_PRECISION"  OFF)
else()
     option(ENABLE_QUAD_PRECISION "Enable compiler definition -DENABLE_QUAD_PRECISION"  ON)
endif()

**How Has This Been Tested?**

Tested as a part of SRW App (short-range weather app) build on Mac OSX (x86_64, M1/arm64).
Other libraries needed for building the SRW App need to be compiled with the QUAD_PRECISION turned off.

**Checklist:**

- [x ] I have performed a self-review of my own code


lharris4:
This should be fine. We are not using ENABLE_QUAD_PRECISION within FV3 as far as I can tell.

bensonr:
@natalie-perlin - is there an inheritance issue or some other reason -DENABLE_QUAD_PRECISION=OFF does not work?

climbfuji:
If I am not mistaken, QUAD_PRECISION is fine on Intel Macs, just not on M1s. You can simply turn off QUAD_PRECISION at the ufs-weather-model top level. You'll need to do that anyway, because stochastic_physics also need to be built w/o quad precision (we implemented that option a couple of months back).

natalie-perlin:
@bensonr - yes, that's what my suggestion was. It could be turned off explicitly in the CMakeLists.exe by setting the option to "OFF", or it could be set to "OFF" when Mac OS is detected:  if(APPLE) .... else()... endif()

climbfuji:
> @bensonr - yes, that's what my suggestion was. It could be turned off explicitly in the CMakeLists.exe by setting the option to "OFF", or it could be set to "OFF" when Mac OS is detected: if(APPLE) .... else()... endif()

I am not convinced that `if(APPLE)` is the right approach, because you can build with `QUAD_PRECISION` on Macs with Intel chips, and even on my Mac M1 when using Rosetta2 `x86_64` emulator mode. Unless I am mixing things up entirely, you only need to turn it off when you run on M1 in native `aarch64` mode. I am happy to try these different options again to confirm.

natalie-perlin:
@climbfuji 

- Yes, it indeed could be compiled with "QUAD_PRECISION" on x86_64, but not on M1/arm64 when running natively (not with Rosetta). Maybe the question is rather whether one would want to run such an application as an atmospheric model with QUAD_PRECISION on a Mac, but things could change quickly these days!..

- It's good to know that CMakeLists.txt is supposed to be configured by the user (from the perspective of writing documentation and assistance to users), and there are other modules/components that need the QUAD_PRECISION=OFF. Compiling on the top level and passing the cmake option such as 
`-DENABLE_QUAD_PRECISION=OFF` does not have any effect, as the 
file CMakeLists.txt in the directory _./src/ufs-weather-model/FV3/atmos_cubed_sphere/_
 is still created with the ENABLE_QUAD_PRECISION set to ON.
 Running cmake as follows from the ./build directory:
 
 ` $ cmake .. -DCMAKE_INSTALL_PREFIX=.. -DENABLE_QUAD_PRECISION=OFF `
 
It also produces a following message:

```
CMake Warning:
  Manually-specified variables were not used by the project:
 
    ENABLE_QUAD_PRECISION
```
 
So I guess there is no need to modify the code, but instead to instruct users on how to change the setting in CMakeLists.txt

PR #196, opened 2022-06-07
fix index error

### LINKED ISSUE #195: Uninitialized variable inf remap_restart
**Describe the bug**
The model crashes when trying to read data from a restart file that has a different number of vertical levels than you
want to use in the run.  More specifically, when `npz .ne. npz_rst`.  The cause of this is the use of the uninitialized variable
`n` used here before it is assigned:

https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/69d97bc1a8025a55b0fe14b27e15c00298f38662/tools/fv_io.F90#L668-L673

**To Reproduce**
Perform a run and set `npz_rst` to a value that is different from `npz` in `fv_core_nml`.

**Expected behavior**
The model should not crash, and be able to remap the restart data to correct number of vertical levels.

**System Environment**
It' caused by undefined compiler behavior, so I'd expect to see it on all systems.

**Additional context**
It can be solved by setting `n = 1`.

menzel-gfdl:
**Description**

Fixes #195 

**How Has This Been Tested?**

This resolved a crash in a run of AM4.0 where I was trying to convert from a c384L33 configuration to a c384L65 configuration.
The run was done on Gaea using the intel 18 compiler suite.

**Checklist:**

### DIFF
--- a/tools/fv_io.F90
+++ b/tools/fv_io.F90
@@ -574,7 +574,7 @@ contains
 
     character(len=64)    :: fname, tracer_name
     character(len=6)     :: stile_name
-    integer              :: isc, iec, jsc, jec, n, nt, nk, ntracers, ntprog, ntdiag
+    integer              :: isc, iec, jsc, jec, nt, nk, ntracers, ntprog, ntdiag
     integer              :: isd, ied, jsd, jed
     integer              :: ntiles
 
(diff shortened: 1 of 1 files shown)


laurenchilutti:
@menzel-gfdl Do you need this fix on the dev/gfdl branch for AM4.0?

PR #206, opened 2022-08-30
Fixes needed for cce compiler

mcallic2:
**Description**
These are the changes that need to be completed to run with the cce compiler 
- some syntax errors fixed
- fixes cray pointer

**How Has This Been Tested?**
Compiled with the cce14 compiler on gaea t5-login1

**Checklist:**

### DIFF
--- a/model/boundary.F90
+++ b/model/boundary.F90
@@ -2214,20 +2214,13 @@ contains
 
    real :: var_nest_3d(is_n:ie_n+istag,js_n:je_n+jstag,1)
    real :: var_coarse_3d(isd_p:ied_p+istag,jsd_p:jed_p+jstag,1)
+   integer( KIND = 8) :: ptr_nest=0
+   integer( KIND = 8) :: ptr_coarse=0
    pointer(ptr_nest, var_nest_3d)
    pointer(ptr_coarse, var_coarse_3d)
 
-   if (child_proc .and. size(var_nest) > 1) then
-           ptr_nest = LOC(var_nest)
-   else
-           ptr_nest = 0
-   endif
-
-   if (parent_proc .and. size(var_coarse) > 1) then
-           ptr_coarse = LOC(var_coarse)
-   else
-           ptr_coarse = 0
-   endif
+   if (child_proc .and. size(var_nest) > 1) ptr_nest = LOC(var_nest)
+   if (parent_proc .and. size(var_coarse) > 1) ptr_coarse = LOC(var_coarse)
 
    call update_coarse_grid_mpp(var_coarse_3d, var_nest_3d, &
         nest_domain, dx, dy, area, &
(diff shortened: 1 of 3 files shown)


model/boundary.F90
```diff
@@ -2214,13 +2214,20 @@ subroutine update_coarse_grid_mpp_2d(var_coarse, var_nest, nest_domain, dx, dy,
...
+           ptr_nest = LOC(var_nest)
+   else
+           ptr_nest = 0
+   endif
+
+   if (parent_proc .and. size(var_coarse) > 1) then
+           ptr_coarse = LOC(var_coarse)
+   else
+           ptr_coarse = 0
+   endif
```
bensonr, line 2230: I'm leery about this modification as you don't have a test case to run it against.  I understand these are Cray pointers and the pointer specification should suffice. but would want to have some test cases run against it.

thomas-robinson: This was probably the most difficult one to change because it's not as straight forward.  The original code looked like this:
```fortran
   integer( KIND = 8) :: ptr_nest=0		
   integer( KIND = 8) :: ptr_coarse=0		
   pointer(ptr_nest, var_nest_3d)		￼   
   pointer(ptr_coarse, var_coarse_3d)		￼
￼
   if (child_proc .and. size(var_nest) > 1) ptr_nest = LOC(var_nest)		￼   
   if (parent_proc .and. size(var_coarse) > 1) ptr_coarse = LOC(var_coarse)
```
It looks like the intent is to set the pointer to 0 and then if the condition is met, the pointer gets the value for the location of the variable.  That's exactly what our code is doing, just without initializing the value to 0 and instead setting it in the if statement.  The crash happened because of 
```fortran
   integer( KIND = 8) :: ptr_nest=0		
   integer( KIND = 8) :: ptr_coarse=0		
```
Is there a test we can run to check that this is working correctly?

bensonr: The original programming contains a gotcha within Fortran that is not well known.  When you give a local variable a value during declaration, it is automatically promoted to include the 'save' attribute.  Upon re-entry, the variable will not get reset to the declaration value, but instead retains the value of the last assignment.

I don't know if the original author understood this or not, but the modified code will not reproduce the original.  I'm not sure why cray pointers, not part of the language standard last time I checked, are used and whether the same behavior could be achieved using fortran pointers.

lharris4:
Hi, Rusty. This implicit 'save' attribute bit us a few months ago regarding
a separate issue in the physics.

I presume that the code could be implemented just as easily using standard
fortran pointers. If I wrote this code (I can't remember if I did) then I
wouldn't have any reason to use the Cray pointers.

Thanks,
Lucas

junwang-noaa:
@bensonr @lharris4 This may not be related. But I want to let you know that we see warnings message from those pointers/LOC statements on wcoss2 when we are asked to remove "nowarn" compile options:

"warning #6843: A dummy argument with an explicit INTENT(OUT) declaration is not given an explicit value."

This may have to be fixed in the next implementation and currently we don't have a solution yet.

bensonr (approved):

thomas-robinson (approved):
The cray pointers still need to be dealt with in the future in order for the cce compiler to work

PR #207, opened 2022-09-02
Implementing CI

laurenchilutti:
**Description**
Parallelworks is currently down, so I will keep this as a draft until it comes back on. - This is resolved.

This PR brings in a github actions yaml file that uses a Self-Hosted Runner (Parallelworks rdhpcs cloud resources) to automatically compile solo shield and run idealized tests on Pull Requests to the main branch of the GFDL_atmos_cubed_sphere repository.  The actual scripts referenced in this yaml are stored on the Parallelworks resource.  The following is a description of how the CI works:

- The CI automatically triggers when anyone submits a PR to the main branch
- The CI consists of 3 stages:
 1. Build
 2. Test
 3. Shut down
- The Build stage has 4 steps:
 1. Checkout Code
 2. Build the sw solo shield configuration
 3. Build the nh solo shield configuration
 4. Build the hydro solo shield configuration
- The Test stage only contains one test in this initial release.  This stage will run a test and then compare restarts against a baseline.  The test will fail if the run fails or the restarts differ
- We are working to speed up these tests on the cloud (currently they are too slow to run all idealized tests)

**How Has This Been Tested?**

This has been tested with generating test PRs in my fork: 

https://github.com/laurenchilutti/GFDL_atmos_cubed_sphere/pull/14 shows that when a PR targets a branch not named "main" the tests will not run

https://github.com/laurenchilutti/GFDL_atmos_cubed_sphere/pull/15 shows that when a PR targets the main branch tests run 

**Checklist:**

(Relies on https://github.com/NOAA-GFDL/SHiELD_build/pull/13)
(Relies on https://github.com/NOAA-GFDL/SHiELD_build/pull/15)

### DIFF
--- a/.github/workflows/SHiELD_parallelworks.yml
+++ b/.github/workflows/SHiELD_parallelworks.yml
@@ -1,0 +1,45 @@
+name: Compile SHiELD SOLO and run tests
+
+on:
+  pull_request:
+    branches:
+      - main
+jobs:
+  buildstep:
+    runs-on: [self-hosted]
+    name: SOLO SHiELD build
+    strategy:
+      fail-fast: true
+      max-parallel: 1
+      matrix:
+        runpath: [/pw/storage/PWscripts]
+        runscript: [FV3checkoutStartClusters.py, FV3swStartClusters.py, FV3nhStartClusters.py, FV3hydroStartClusters.py]
+    steps:
+      - env:
+          RUNPATH: ${{ matrix.runpath }}
+          RUNSCRIPT: ${{ matrix.runscript }}
+        run: python3 $RUNPATH/$RUNSCRIPT $GITHUB_REF
+        
+  tests:
+    runs-on: [self-hosted]
+    name: SOLO SHiELD test suite
+    needs: [buildstep]
+    strategy:
+      fail-fast: false
+      max-parallel: 3
+      matrix:
+        runpath: [/pw/storage/PWscripts]
+        runscript: [FV3C128r20.solo.superCStartClusters.py]
+    steps:
+      - env:
+          RUNPATH: ${{ matrix.runpath }}
+          RUNSCRIPT: ${{ matrix.runscript }}
+        run: python3 $RUNPATH/$RUNSCRIPT $GITHUB_REF
+        
+  shutdowncluster:
+    runs-on: [self-hosted]
+    name: Shutdown cluster
+    if: always()
+    needs: [buildstep, tests]
+    steps:
+      - run: python3 /home/Lauren.Chilutti/pw/storage/PWscripts/stopClusters.py ci_fv3
--- a/.github/workflows/SHiELD_parallelworks.yml
+++ b/.github/workflows/SHiELD_parallelworks.yml
@@ -5,25 +5,32 @@ on:
     branches:
       - main
 jobs:
-  buildstep:
+  checkout:
+    runs-on: [self-hosted]
+    name: Checkout Code
+    steps:
+    - run: python3 /pw/storage/PWscripts/FV3checkoutStartClusters.py $GITHUB_REF
+    
+  build:
     runs-on: [self-hosted]
     name: SOLO SHiELD build
+    needs: [checkout]
     strategy:
       fail-fast: true
-      max-parallel: 1
+      max-parallel: 3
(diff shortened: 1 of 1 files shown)


.github/workflows/SHiELD_parallelworks.yml
```diff
@@ -0,0 +1,45 @@
...
+  buildstep:
+    runs-on: [self-hosted]
+    name: SOLO SHiELD build
+    strategy:
+      fail-fast: true
+      max-parallel: 1
+      matrix:
+        runpath: [/pw/storage/PWscripts]
+        runscript: [FV3checkoutStartClusters.py, FV3swStartClusters.py, FV3nhStartClusters.py, FV3hydroStartClusters.py]
+    steps:
```
thomas-robinson, line 17: Do each of these run through the cluster spin up part of the API script?

.github/workflows/SHiELD_parallelworks.yml
```diff
@@ -0,0 +1,45 @@
...
+  pull_request:
+    branches:
+      - main
+jobs:
+  buildstep:
+    runs-on: [self-hosted]
+    name: SOLO SHiELD build
+    strategy:
+      fail-fast: true
+      max-parallel: 1
```
thomas-robinson, line 13: How many runners do you have?  You have 4 jobs, do they need to all run one after the other?  You could set up 4 runners to run all 4 at the same time.

.github/workflows/SHiELD_parallelworks.yml
laurenchilutti: They all run through the start up script and just have a different testcmd

laurenchilutti: The first only checks out code, the next 3 each build a different configuration

.github/workflows/SHiELD_parallelworks.yml
laurenchilutti: I only have 1 right now.  But I will set up multiple runners so that we can do multiple tests at the same time.  For the build step I need them all to build in series (and not parallel).

bensonr: I believe the builds could be done in parallel if we have the CI system build the FMS library first.

laurenchilutti: Would there be an issue with the builds occurring simultaneously when they all use the same Build/exec folder? I could alter the Build system so that they all build with different exec folders

bensonr: The current build system uses exec/${config}_${compiler}/  It could be altered to use a combination of _config, hydro, comp, bit, and compiler_

laurenchilutti: I will start making those changes - building in parallel would make everything much faster.

bensonr: Remember to run the FMS build as a standalone process.  Once we bring in the updates to build the few NCEPlibs needed by SHiELD, that would become part of that first step.

laurenchilutti:
I have created PR https://github.com/NOAA-GFDL/SHiELD_build/pull/15 to now have scripts to build libFMS and nceplibs.  There is also the ability to store these libraries external to the SHiELD_build so that libFMS does not need to be built with every new CI instance.  And I have updated the compile such that there are unique exec/ sub-directories for each configuration so that we can do all 3 builds in parallel.  

I have also added more Github runners (3 in total) so we can run builds and tests in parallel.  Once This PR from SHiELD_build is merged I will rerun the CI on this PR so we can see how it works

bensonr (approved):

thomas-robinson:
@laurenchilutti After this gets merged it, submit a PR with code that should fail to make sure that the tests fail when expected.

laurenchilutti:
Tests passed and builds ran in parallel - future improvements I want to make: 
- Add more tests
- Add more runners if needed (currently there are 3 so we can only run 3 jobs in parallel)
- Make it such that any nh configuration tests only rely on the completion of the nhcompile, and the same for hydro and sw
- Add tests that compare different layouts
- Add tests that run from restarts to ensure restart functionality is as expected.

Issue #212, opened 2022-09-20
no pressure data on velocity points for remap dwinds


SamuelTrahanNOAA:
**Describe the bug**

The remap_dwinds needs surface pressure ("ps" in boundary files) to remap the winds. The surface pressure is not available on the outermost velocity point, so the resulting winds are nonsense.

**To Reproduce**
Initialize ps_reg to real_snan in regional_bc_data and witness the crash

**Expected behavior**
Reasonable pressure should be available on all velocity points.

SamuelTrahanNOAA:
Will open new issue shortly.

PR #216, opened 2022-09-20
Fixes many issues in fv_regional_bc.F90 and allows debug & 2threads RRFS variants to pass

### LINKED ISSUE #213: remap_scalar_nggps_regional_bc overwrites Atm%ps for entire grid, not just boundary points
**Describe the bug**

The remap_scalar_nggps_regional_bc overwrites Atm%ps for the entire grid, for each boundary. This results in gibberish pressure values, especially because most of psc and ps are -9999999. (I'll have another issue on that shortly.) Because of this, the RRFS crashes when run in debug mode.

```
      is=lbound(Atm%ps,1)                                                                                                                                                                                                                   
      ie=ubound(Atm%ps,1)                                                                                                                                                                                                                   
      js=lbound(Atm%ps,2)                                                                                                                                                                                                                   
      je=ubound(Atm%ps,2)                                                                                                                                                                                                                   
!                                                                                                                                                                                                                                            
      do j=js,je
      do i=is,ie
        Atm%ps(i,j)=ps(i,j)
      enddo
      enddo
```

**To Reproduce**
Read the code. Or, run it in debug mode, and examine the core dump using gdb.

**Expected behavior**
Only boundary regions are filled with boundary surface pressure.

### LINKED ISSUE #214: remap_scalar_nggps_regional_bc and remap_dwinds_regional_bc access uninitialized memory
**Describe the bug**
The remap_scalar_nggps_regional_bc does a two point average of `psc(i,j)` and `psc(i,j-1)` (and likewise with i-1,j). This is a problem for three reasons:

1. South boundary is processed before east and west, so it'll access areas the east and west boundaries have not initialized.
2. At edges of the patch, the code will access data in the prior patch (to j-1 or i-1 direction) which are not filled in by that patch due to lack of halos and inter-process communication.
3. Beyond the edge of the boundary region, the `i-1` and `j-1` regions are outside the regions where boundary conditions provide surface pressure.

The reason this does not crash is that ps_reg is initialized with -9999999, detailed in issue #215.

**To Reproduce**
Read the code, or initialize ps_reg to real_snan, run in debug mode and examine the core dump using gdb

**Expected behavior**
Code should not do an average of valid data with gibberish.

### LINKED ISSUE #215: intentional ps_reg=-9999999 bug in fv_regional_bc.F90 to hide lack of testing
**Describe the bug**

After developing some very buggy software (#212, #213, #214) instead of testing it and fixing the bugs, the developer hid the issues by initializing surface pressure to -9999999. Due to several other errors (not caught due to the lack of testing), this propagated to the model, causing floating point exceptions when the model is run in debug mode with boundary conditions. Thanks to this intentional bug, the RRFS has not been tested in debug mode for a long time. In addition, there are gibberish values in all of the fields at the edge of processors and at the edge of the boundary due to interpolation with -9999999.

In fv_regional_bc.F90:

```
  allocate(ps_reg(is_input:ie_input,js_input:je_input)) ; ps_reg=-9999999 ! for now don't set to snan until remap dwinds is changed                                                                                                   
```

It should be noted that the problems are not restricted to "remap dwinds" but are throughout fv_regional_bc.F90 and even in the boundary condition files themselves, which do not provide "ps" on all wind points. Had this code been tested, instead of hidden behind a -9999999, these bugs would have been caught and fixed a long time ago.

**To Reproduce**
Read the code. Or, turn on debug mode and witness -9999999 showing up later in the model.

**Expected behavior**
This should never have been in a pull request, and certainly should not have passed review.

SamuelTrahanNOAA:
**Description**

Initializes ps_reg to real_snan (signalling NaN) and corrects all issues that come along later. This allows the RRFS 13km conus domain regression tests to run in debug and 2threads mode. It should fix gibberish boundary conditions and dubious initial surface pressures seen throughout the domain when boundary conditions are enabled.

I'd like to stress that the reason this PR is needed, is that a developer hid these bugs by initializing surface pressure ("ps_reg" in fv_regional_bc.F90) to -9999999 instead of debugging and fixing their code. That -9999999 propagates throughout fv_regional_bc.F90 and elsewhere in the model, adversely affecting many things. (See #215 for details.) Although this allows the model to run without crashing when debug mode is disabled, it harms the forecast in many ways. It also prevented us from testing the RRFS in debug mode for a long time.

Fixes #213
Fixes #214 
Fixes #215

This is a workaround for #212, but it is NOT a fix. The bad boundary files have to be fixed in the program that generates those files.

**How Has This Been Tested?**

Hera and Jet intel tests of the RRFS conus 13km domain in debug mode.

**Checklist:**

- [x] I have made corresponding changes to the documentation (no changes needed to documentation)
- [x] Any dependent changes have been merged and published in downstream modules (no dependencies)

### DIFF
--- a/model/fv_fill.F90
+++ b/model/fv_fill.F90
@@ -37,7 +37,7 @@ module fv_fill_mod
 ! </table>
 
    use mpp_domains_mod,     only: mpp_update_domains, domain2D
-   use platform_mod,        only: kind_phys => r8_kind
+   use GFS_typedefs,        only: kind_phys
 
    implicit none
    public fillz
--- a/model/fv_regional_bc.F90
+++ b/model/fv_regional_bc.F90
@@ -1913,42 +1913,21 @@ contains
         allocate(ps_reg(is_input:ie_input,js_input:je_input))              !<-- Sfc pressure in domain's boundary region derived from BC files
 !
 !-----------------------------------------------------------------------
-!*** Whoever did this intentionally added a bug to the code and hid it.
-!*** I'm retaining this code, commented out, as a reminder to others.
+!*** Whoever did this intentionally added a bug to the code and hid it,
+!*** to avoid fixing other bugs elsewhere.
+!*** The result is that RRFS debug mode was broken for a long time
+!*** and some boundary data was gibberish.
+!*** I'm retaining this code, commented out, as a reminder to others
+!*** of what NOT to do.
 !-----------------------------------------------------------------------
 !
         !ps_reg=-9999999 ! for now don't set to snan until remap dwinds is changed
+!
+!-----------------------------------------------------------------------
+!*** Initialize ps_reg to real_snan to detect errors later in code.
+!-----------------------------------------------------------------------
+!
         ps_reg=real_snan
-!
-!-----------------------------------------------------------------------
-!*** Initialize the domain boundary surface pressure to the input
-!*** surface pressure on points remap dwinds can't handle.
-!*** When data is missing or invalid, use nearby points. If all nearby
-!*** points are invalid, use standard atmospheric pressure.
-!-----------------------------------------------------------------------
-!
-        do j=js_input,je_input
-          do i=is_input,ie_input
-            if(ieee_is_finite(ps_input(i,j,1))) then
-              ps_reg(i,j) = ps_input(i,j,1)
-
-            ! When there is invalid data at a point, search nearby points.
-            else if(ieee_is_finite(ps_input(max(is_input,i-1),j,1))) then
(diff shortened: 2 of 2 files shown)


SamuelTrahanNOAA:
The ufs-weather-model develop and atmos_cubed_sphere dev/emc updated since my last test, so I'm retesting before making the ufs-weather-model PR.

SamuelTrahanNOAA:
I'm running into this bug in gfortran:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=82207

It is fixed in gcc 12, but that is not a viable option.

SamuelTrahanNOAA:
Will submit new PR shortly, after more testing.

PR #219, opened 2022-09-22
Fix several bugs in fv_regional_bc.F90 relating to uninitialized or incorrectly initialized memory.

### LINKED ISSUE #218: uninitialized or incorrectly initialized memory in fv_regional_bc.F90
**Describe the bug**

The ps_reg variable in fv_regional_bc.F90 is initialized to -9999999 Pa, which propagates throughout the model. All other boundary arrays were initialized to signalling NaN. Not initializing this one to signalling NaN masked many other bugs. 

These are some of the bugs it masked inside the GFDL_atmos_cubed_sphere repository; there are others outside it:

1. Remapping winds does a two point average of the current point and -9999999.
2. Remapping winds uses -9999999 as pressure on velocity boundary points beyond the scalar boundary points.
3. The -9999999 is copied to Atm%ps throughout the domain, beyond the boundary, on all processors that touch the domain boundary.
4. Due to item 3, the RRFS hasn't been able to run in DEBUG=ON mode since this bug was introduced.
5. The pressure data is not read in on velocity points outside the scalar boundary.  (It can't be, since the the boundary condition files lack that data entirely.)
6. The blending code blends data beyond boundary regions.
7. The code accesses i-1 and j-1 indices, but processes the south boundary before the east and west boundary. That means two south boundary processes are accessing uninitialized regions instead of the east and west boundary.

There are other issues, outside the dynamical core, which were masked by this bug. I'll link to those issues once I create or find them.

- https://github.com/ufs-community/ufs-weather-model/issues/1222
- https://github.com/ufs-community/ufs-weather-model/issues/1436
- https://github.com/ufs-community/ccpp-physics/issues/5
- https://github.com/NOAA-EMC/fv3atm/issues/586

It should also be noted that many users have found dubious values the model receives from the boundary conditions. These bugs may be the cause. However, we'll have to run parallels of multiple configurations to confirm that.

**To Reproduce**

1. Read the code and find the ps_reg=-9999999
2. Replace that with ps_reg=real_snan
3. Compile in DEBUG=ON mode.
4. Run it and see the bugs.
6. Fix the bugs one by one and see more appear.

**Expected behavior**

1. Boundary data should be correct.
2. The boundary code should not write to non-boundary regions within the domain.
3. Blending code should not blend data beyond the domain boundary.
4. The best practice of initializing to real_snan should be followed.
5. The code should pass in DEBUG=ON
8. Changing the threads, layout, or task count, should not change the results.

**System Environment**
The ufs-weather-model build system targets hera.intel and hera.gnu, which use:

- OS: CentOS 7.9.2009
- Compiler(s): Intel 2022.1.2 and GCC 9.2.0
- MPI type, and version: Intel MPI using the SLURM srun launcher
- netCDF Version: 4.7.4
- Configure options: cmake option "-DDEBUG=ON"

**Additional context**
None.

SamuelTrahanNOAA:
**Description**

This removes the workaround of initializing ps_reg to -9999999 Pa in fv_regional_bc.F90 and fixes many other bugs that caused or masked. One bug related to blending has a workaround instead of a fix. A lot of these bugs were fixed by merging in #220. That PR from @kaiyuan-cheng has an elegant replacement for some of my earlier kludgy fixes. Unfortunately, that PR does not allow the model to run in debug mode. I believe this PR has the minimal set of fixes and workarounds needed, until someone comes up with a more elegant solution to the blending bug.

**What is Fixed?**

This PR fixes everything listed here, while related PRs (listed below) at higher levels fix several other issues. See the next section for the one bug that has a workaround instead of a fix.

1. Only valid boundary values are copied to Atm%ps (not to data within the domain)
2. Many uninitialized variables are now initialized with signalling NaN (real_snan)
3. ps_reg is no longer initialzied to -9999999 Pa
4. Replaced "go to 123" with "exit"
5. PR #220 is included. Indexes are corrected so that wind remapping never accesses data outside boundary regions.

**Workaround for Blending Code Receiving Data Outside the Boundary**

The blending code's `array` array contains NaNs in regions it is asked to blend. If my reading of the code is correct, the `array` array is what is being blended, not the boundary conditions themselves. This is probably (but not certainly) regions beyond the boundary in the domain. Any operations with a NaN return NaN (unless the NaN signals), so processing those locations is a no-op (or a SIGFPE). The workaround is to detect non-finite values using `is_not_finite()` and skip them.

I'm concerned by this bug, as it means something may be wrong elsewhere. Someone with better knowledge of the code should look into this. The trouble may originate from the model level, so I've made an issue there (https://github.com/NOAA-EMC/fv3atm/issues/586). It may be as simple as the blending code looping one point farther than it should.

**Testing:**

Fixes #218 

The RRFS regression tests in the ufs-weather-model were compiled in DEBUG=ON mode, and extensively tested on Hera with the Gnu and Intel compilers. Many hours of debugging with gdb, ncview, and other tools. Testing this also required fixing a numerical precision issue in one ccpp physics scheme (https://github.com/ufs-community/ccpp-physics/issues/5, https://github.com/ufs-community/ccpp-physics/pull/6) and changes to the regression tests to get them to fit in 30 minutes (https://github.com/ufs-community/ufs-weather-model/pull/1437). Ultimately, this fixes half of the test variants of RRFS that were failing before (https://github.com/ufs-community/ufs-weather-model/issues/1222). It also obviates the problem of not having enough PS boundary data (https://github.com/ufs-community/ufs-weather-model/issues/1436).

Preliminary results from a real-time parallel have shown this has a small impact on the bulk 1D statistics in a large domain with a large number of ranks. (Hence, boundary ranks have a small fraction of the domain.) However, regression tests have shown it does change the fields. Nobody has done any analysis between those two extremes.

**Checklist:**

### DIFF
--- a/model/fv_regional_bc.F90
+++ b/model/fv_regional_bc.F90
@@ -235,42 +235,12 @@ module fv_regional_mod
       character(len=100) :: grid_data='grid.tile7.halo4.nc'             &
                            ,oro_data ='oro_data.tile7.halo4.nc'
 
-!
-!-----------------------------------------------------------------------
-!***  All arrays and variables that contain boundary data should be
-!***  initialized to real_snan or dbl_snan (as appropriate) because that
-!***  is the only way to detect errors in the boundary generation.
-!***  Usual approaches like detecting out-of-bounds accesses won't work
-!***  because the arrays may be larger than the boundaries.
-!***  
-!***  Regions of the arrays that should have valid boundary data should
-!***  never contain a NaN. If they do, then there is an error that has
-!***  not yet been identified. In particular, the blending code has an
-!***  unidentified problem.
-!***
-!***  Presently, the blending code is asked to blend data in regions
-!***  that contain NaN. This means either:
-!***
-!***  1) not all boundary areas are filled in, OR
-!***  2) the blending code is accessing beyond the boundaries
-!***
-!***  The reasons for this are unknown, but the workaround does work:
-!***  detect the NaNs and don't use those points in blending.
-!***
-!***  Also, the wind remapping code needs surface pressure beyond the
-!***  wind boundary regions, but the surface pressure boundary data
-!***  doesn't extend that far.
-!-----------------------------------------------------------------------
-!
 #ifdef OVERLOAD_R4
-      real, parameter:: real_snan=real(Z'FFBFFFFF') ! <-- One of many ways to express 32-bit signalling NaN
-      real, parameter:: real_qnan=real(Z'FFFFFFFF') ! <-- One of many ways to express 32-bit quiet NaN
+      real, parameter:: real_snan=real(Z'FFBFFFFF')
 #else
-      real, parameter:: real_snan=real(Z'FFF7FFFFFFFFFFFF') ! <-- One of many ways to express 64-bit signalling NaN
-      real, parameter:: real_qnan=real(Z'FFFFFFFFFFFFFFFF') ! <-- One of many ways to express 64-bit quiet NaN
+      real, parameter:: real_snan=real(Z'FFF7FFFFFFFFFFFF')
 #endif
-      real(kind=R_GRID), parameter:: dbl_snan=real(Z'FFF7FFFFFFFFFFFF',kind=R_GRID) ! <-- One of many ways to express 64-bit signalling NaN
-      real(kind=R_GRID), parameter:: dbl_qnan=real(Z'FFFFFFFFFFFFFFFF',kind=R_GRID) ! <-- One of many ways to express 64-bit quiet NaN
+      real(kind=R_GRID), parameter:: dbl_snan=real(Z'FFF7FFFFFFFFFFFF',kind=R_GRID)
 
       interface dump_field
         module procedure dump_field_3d
@@ -283,6 +253,7 @@ module fv_regional_mod
 !
       logical :: data_source_fv3gfs
 contains
+
 
 !-----------------------------------------------------------------------
 !
@@ -306,7 +277,7 @@ contains
 ! supported in older compilers.
 !-----------------------------------------------------------------------
 !
-    intrinsic shiftr, transfer, iand ! <--- for the benefit of older compilers
+    intrinsic shiftr, transfer, iand ! <--- declare intrinsic to help older compilers
 !
 !-----------------------------------------------------------------------
 ! Use value-based argument passing instead of reference-based to avoid
@@ -344,8 +315,6 @@ contains
 !
   end function is_not_finite
 !
-!-----------------------------------------------------------------------
-
 !-----------------------------------------------------------------------
 !
       subroutine setup_regional_BC(Atm                                  &
@@ -1043,7 +1012,7 @@ contains
         regional_bc_bounds%ie_north_uvs=ied
 !
         regional_bc_bounds%js_north_uvs=jsd
-        regional_bc_bounds%je_north_uvs=nrows_blend+1
+        regional_bc_bounds%je_north_uvs=nrows_blend
 !
         regional_bc_bounds%is_north_uvw=isd
         regional_bc_bounds%ie_north_uvw=ied+1
@@ -1060,7 +1029,7 @@ contains
         regional_bc_bounds%is_south_uvs=isd
         regional_bc_bounds%ie_south_uvs=ied
 !
-        regional_bc_bounds%js_south_uvs=jed-nhalo_model-nrows_blend+1
+        regional_bc_bounds%js_south_uvs=jed-nhalo_model-nrows_blend+2
         regional_bc_bounds%je_south_uvs=jed+1
 !
         regional_bc_bounds%is_south_uvw=isd
@@ -1120,7 +1089,7 @@ contains
           regional_bc_bounds%je_west_uvs=jed-nhalo_model+1
         endif
 !
-        regional_bc_bounds%is_west_uvw=ied-nhalo_model-nrows_blend+1
+        regional_bc_bounds%is_west_uvw=ied-nhalo_model-nrows_blend+2
         regional_bc_bounds%ie_west_uvw=ied+1
 !
         regional_bc_bounds%js_west_uvw=jsd
@@ -2035,7 +2004,16 @@ contains
           endif
         endif
 !
-          if(nside==2)then
+        if(nside==2)then
+          if(south_bc)then
+            call_remap=.true.
+            side='south'
+            bc_side_t1=>BC_t1%south
+            bc_side_t0=>BC_t0%south
+          endif
+        endif
+!
+          if(nside==3)then
             if(east_bc)then
               call_remap=.true.
               side='east '
@@ -2044,21 +2022,12 @@ contains
             endif
           endif
 !
-        if(nside==3)then
+        if(nside==4)then
           if(west_bc)then
             call_remap=.true.
             side='west '
             bc_side_t1=>BC_t1%west
             bc_side_t0=>BC_t0%west
-          endif
-        endif
-!
-        if(nside==4)then
-          if(south_bc)then
-            call_remap=.true.
-            side='south'
-            bc_side_t1=>BC_t1%south
-            bc_side_t0=>BC_t0%south
           endif
         endif
 !
@@ -2138,59 +2107,7 @@ contains
               endif
            endif
 
-!
-
            if(nside==2)then
-              if(east_bc)then
-                if (js == 1) then
-                    jstart = 1
-                else
-                    jstart = jsd
-                endif
-                if (je == npy-1) then
-                   jend = je
-                else
-                   jend = jed
-                endif
-
-
-                do k=1,npz
-                do j=jstart,jend
-                do i=isd,0
-                     delz_regBC%west_t1(i,j,k) = bc_side_t1%delz_BC(i,j,k)
-                     delz_regBC%west_t0(i,j,k) = bc_side_t0%delz_BC(i,j,k)
-                enddo
-                enddo
-                enddo
-              endif
-            endif
-
-          if(nside==3)then
-              if(west_bc)then
-                if (js == 1) then
-                    jstart = 1
-                else
-                    jstart = jsd
-                endif
-                if (je == npy-1) then
-                   jend = je
-                else
-                   jend = jed
-                endif
-
-
-                do k=1,npz
-                do j=jstart,jend
-                do i=npx,ied
-                     delz_regBC%east_t1(i,j,k) = bc_side_t1%delz_BC(i,j,k)
-                     delz_regBC%east_t0(i,j,k) = bc_side_t0%delz_BC(i,j,k)
-                enddo
-                enddo
-                enddo
-              endif
-            endif
-
-           if(nside==4)then
               if(south_bc)then
                 if (is == 1) then
                    istart = 1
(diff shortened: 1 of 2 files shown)


SamuelTrahanNOAA:
These changes are going into the RRFS as an emergency fix, in this PR:

https://github.com/NOAA-GSL/GFDL_atmos_cubed_sphere/pull/12

It's headed to the RRFS_dev branch in the NOAA-GSL fork, which is used as the authoritative repo for RRFS parallels. I'm hoping to un-fork the GFDL_atmos_cubed_sphere as soon as possible. The developers do not want wait the several weeks it'll take to update RRFS_dev to the authoritative branches. They want this fix as soon as possible, due to the severity of the bug and its widespread implications.

junwang-noaa:
@yangfanglin Could we ask Kate or Kevin to review this PR? Thanks

laurenchilutti:
Thank you Jun, I wasn't sure who to ask to review this.  I can update the reviewers if you would like.

junwang-noaa:
Thanks, Lauren. I think some dynamics export on UFS regional application can review the PR.

kaiyuan-cheng:
The original code looks good to me. The proposed change does not seem correct because it remaps the winds at cell edges based on surface pressure at the cell center. FV3 needs winds on the cell edges (C- and D-grids). When the model runs in the regional mode, the remapping of the winds in the boundary domain requires the surface pressure to be on the cell edges, which is obtained by the linear interpolation of the surface pressure at the cell center. For example, 0.5*(psc(i,j-1)+psc(i,j)), as shown in the original code. I am uncertain if there is an issue with the boundary condition files. There are four halos of data, which should be more than enough to construct winds at the edges of three halos required by the FV3 algorithm. For this reason, I don’t see when the lack of surface pressure for remapping winds will happen.

model/fv_regional_bc.F90
```diff
@@ -3975,12 +4214,12 @@ subroutine remap_dwinds_regional_bc(Atm                              &
 
      do k=1,km+1
         do i=is_v,ie_v
-           pe0(i,k) = ak0(k) + bk0(k)*0.5*(psc(i-1,j)+psc(i,j))
+           pe0(i,k) = ak0(k) + bk0(k)*psc(i,j)
```
bensonr, line 4217: I have the same question as @kaiyuan-cheng - why are you removing the average ,which provides the pressure at the cell edge values where the winds are defined, in favor of the cell mean value defined at the cell center?

model/fv_regional_bc.F90
```diff
@@ -3975,12 +4214,12 @@ subroutine remap_dwinds_regional_bc(Atm                              &
...
         enddo
      enddo
      do k=1,npz+1
         do i=is_v,ie_v
-           pe1(i,k) = Atm%ak(k) + Atm%bk(k)*0.5*(psc(i-1,j)+psc(i,j))
+           pe1(i,k) = Atm%ak(k) + Atm%bk(k)*psc(i,j)
```
bensonr, line 4222: Same as the previous comment.

model/fv_regional_bc.F90
SamuelTrahanNOAA: The i-1 value is outside the boundary in some places.

model/fv_regional_bc.F90
SamuelTrahanNOAA: The j-1 value is outside the boundary in some places.

SamuelTrahanNOAA:
> For this reason, I don’t see when the lack of surface pressure for remapping winds will happen.

Turn back on ps_reg=real_snan and you will see real_snan in those two lines and quite a few others. The code extensively uses uninitialized memory.

SamuelTrahanNOAA:
There are three reasons I know of why surface pressure is unavailable in the wind remapping:

1. PS is unavailable in the file on some velocity points (it is only available on scalar value points).
2. The boundaries are processed in the wrong order, so the j-1 points are unavailable.
3. Some of the i-1 and j-1 locations are outside the boundary region.

However, I don't even need to know that any of those are true, to know the code reads uninitialized areas. Just turn on the safeguard ps_reg=real_snan, and you'll see it fail all over the place.

SamuelTrahanNOAA:
The only way I see to put the i-1 and j-1 back in, is to use i-1 or j-1 when they are available, but only the local point when they're not. I can use my "is_not_finite" subroutine to determine that.

There are points where i-1 or j-1 are inside the domain, where pressure is unavailable, or beyond the boundary regions, where pressure is unavailable. We'll need PS in the boundary files to extend into those regions. I'll update my PS-related issue accordingly.

Also, the "PS" reader will have to read into the next MPI rank's boundary region, to get i-1 and j-1.

model/fv_regional_bc.F90
SamuelTrahanNOAA: You will find this same issue for j-1 a few lines up, in two places.

SamuelTrahanNOAA:
As requested by @kaiyuan-cheng and @bensonr the i-1 and j-1 two-point averages are back in the code, and I've used a new workaround:

1. The i-1 and j-1 are used when available
2. No averaging is done with i-1 and j-1 are unavailable
3. Pressure from the nearest scalar point is used when the i,j location is also unavailable.

We cannot fix this properly until https://github.com/ufs-community/ufs-weather-model/issues/1436 is corrected. However, this is a better workaround than using a pressure of -9999999 Pa.

model/fv_regional_bc.F90
SamuelTrahanNOAA: This error has been corrected, along with the similar errors in three other places.

JiliDong-NOAA:
@MatthewPyle-NOAA could you look into this PR? it will be great if we can figure out if there is any potential issues on blending  NaN values at the lateral boundary.

MatthewPyle-NOAA:
Yes @JiliDong-NOAA , but probably won't have time to look at it until early next week.

JacobCarley-NOAA:
Unsure if helpful, but the limited area model approach/design is covered in [Section 2 and Fig. 1 of this paper](https://doi.org/10.1029/2021MS002483)

And I've attached an old description of how the blending was coded from an internal memo.
[Blending.pdf](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/files/9659171/Blending.pdf)

SamuelTrahanNOAA:
I see this in the paper:

> because remapping of the wind components requires adjacent pressure values, a fourth column/row of boundary grid cells is required to hold pressure

Some points lack the j-1 or i-1 pressure data. I see one place where that *will* happen, according to the figure1 in the paper: when i-1 or j-1 are inside the domain. 

However, I think some of the boundaries may be missing that fourth column/row as well, once it gets to the fv_regional_bc.F90. When I try to extend the region that PS is read in, the NetCDF library tells me that is beyond the bounds of the array. Either the data isn't there, or there's some indexing oddity for PS that the code isn't accounting for.

SamuelTrahanNOAA:
@jkbk2004 I'd like to delay this PR until it has been properly reviewed. It is https://github.com/ufs-community/ufs-weather-model/pull/1437 in the ufs-weather-model repository. Could you please reschedule that PR to next week?

kaiyuan-cheng (commented):
I went through the code carefully. It turns out that remap_dwinds_regional_bc() does access uninitialized surface pressure, as @SamuelTrahanNOAA pointed out. It means that winds at certain columns and rows are incorrectly remapped. The good news is that those bad winds are never used by the FV3 dycore. (It is not surprising as we don't observe obvious noise near the boundary.) Although this issue does not affect the solution, a fix would be still desirable. I have the fix ready and I will open a PR shortly.

SamuelTrahanNOAA:
Everyone is hyperfocusing on four lines of code when there are quite a few other bugs in the fv_regional_bc.F90. In particular, the bug causing the model to abort in debug mode is NOT the one involving accessing i-1 and j-1. Since some people don't want to read my bug reports or pull request, I'll give you a hint: -9999999 Pa does propagate up to the model, inside the domain, not just at the boundary.

kaiyuan-cheng:
I do like the idea "ps_reg = real_snan" and I have it in my fix. It does not signal any error in my test case. I am not sure about what other bugs you are referring to.

SamuelTrahanNOAA:
> I do like the idea "ps_reg = real_snan" and I have it in my fix. It does not signal any error in my test case. I am not sure about what other bugs you are referring to.

Good to hear. I will have to see your changes to know what (if anything) is not fixed.

SamuelTrahanNOAA:
@kaiyuan-cheng 's PR is here: #220, but it is on top of main rather than dev/emc. I lack the ability to test this in UFS since it won't compile within UFS. (CMakeFile is missing.) That means I can't even confirm #220 will fix the bugs. It also leaves me wondering if some of the bugs originate from elsewhere in dev/emc, rather than the three boundary index lines fixed in #220. Without the ability to test, I do not know.

SamuelTrahanNOAA:
Although #220 does not fix the problem, I was able to dramatically reduce the complexity of my PR by replacing nearly all of my PS-related fixes with @kaiyuan-cheng's change. I believe this is the minimal possible set of fixes until someone has a more elegant fix to the blending bug.

BrianCurtis-NOAA:
The UFSWM PR that contains this is on the top of the commit queue. I recommend we get reviews/approvals here that there are no visible issues in the code **before** we start testing in UFSWM.

SamuelTrahanNOAA:
I concur with @BrianCurtis-NOAA. I'm less worried about the state of this PR after the fixes from #220 were merged, but I do want someone to vet the remaining changes before widespread testing.

model/fv_regional_bc.F90
```diff
@@ -3594,10 +3663,10 @@ subroutine remap_scalar_nggps_regional_bc(Atm                         &
...
 !---------------------------------------------------------------------------------
 !
-      is=lbound(Atm%ps,1)
-      ie=ubound(Atm%ps,1)
-      js=lbound(Atm%ps,2)
-      je=ubound(Atm%ps,2)
+      is=min(ubound(Atm%ps,1),max(lbound(Atm%ps,1),is))
+      ie=min(ubound(Atm%ps,1),max(lbound(Atm%ps,1),ie))
+      js=min(ubound(Atm%ps,2),max(lbound(Atm%ps,2),js))
+      je=min(ubound(Atm%ps,2),max(lbound(Atm%ps,2),je))
```
kaiyuan-cheng, line 3669: I doubt that the modification here is really necessary.

model/fv_regional_bc.F90
```diff
@@ -4864,6 +4933,9 @@ subroutine bc_time_interpolation(array                               &
           do j=j1_blend,j2_blend
             factor_dist=exp(-(blend_exp1+blend_exp2*(j-j_bc-1)*rdenom)) !<-- Exponential falloff of blending weights.
             do i=i1_blend,i2_blend
+              if(is_not_finite(array(i,j,k))) then
+                cycle ! Outside boundary
+              endif
```
kaiyuan-cheng, line 4938: I don’t run the blending boundary condition so I cannot test it. This walkaround looks fine to me although I do think that revising the indices of the blending domain is the ultimate fix.

SamuelTrahanNOAA: You're probably right about the indices, but I don't trust my knowledge of the code. If someone else knows the blending code, hopefully they'll review and suggest a better solution.

MatthewPyle-NOAA:
I haven't had time to look in great detail, but philosophically it seems like a bandaid is being applied with these is_not_finite checks.  It would be nice to know the root cause of why it is reaching out beyond defined data.   How frequently are these checks excluding i,j values?              

```
if(is_not_finite(array(i,j,k))) then
                cycle ! Outside boundary
endif
```
The other change that caught my eye was the changes to the regional_bc_bounds variables (for south, north, and west) - I just would be somewhat surprised that it had been wrong all this time given the testing and vetting of the code by the original coders.   But certainly could be possible.

```
-        regional_bc_bounds%is_west_uvw=ied-nhalo_model-nrows_blend+1
+        regional_bc_bounds%is_west_uvw=ied-nhalo_model-nrows_blend+2
```

model/fv_regional_bc.F90
SamuelTrahanNOAA: Without these changes, the model uses data outside the boundary regions that are overwritten with real_snan in the following loop. This error was how I found the ps_reg=-9999999 and later bugs. (Maybe this is caused by a region of the code you don't typically use?)

However, even if we assume that the changes are not necessary, it is then a harmless bounds check on copying to Atm%ps. As with initializing data to real_snan, bounds checks are reasonable safeguard.

SamuelTrahanNOAA:
@MatthewPyle-NOAA Those index changes come from @kaiyuan-cheng in #220 and they do prevent most out-of-bounds access. However, the results are different than my original fixes in RRFS_dev so these have not been vetted in a parallel yet.

SamuelTrahanNOAA:
> How frequently are these checks excluding i,j values?

I have not tested that, but I recall in earlier testing the first variable that has the problem is "pt". Also, those changes are a no-op; the areas that it is blending have no valid data. 

> I just would be somewhat surprised that it had been wrong all this time given the testing and vetting of the code by the original coders.

Many out-of-bounds bugs were masked by `ps_reg=-9999999` which bypassed the normal safeguard of initializing to signalling NaN. My issue #218 lists the out-of-bounds accesses I found when restoring the safeguard.

jkbk2004:
All tests for ufs-wm [pr #1437](https://github.com/ufs-community/ufs-weather-model/pull/1437) are finished. We can start merging in this pr.

jkbk2004:
@laurenchilutti @SamuelTrahanNOAA @bensonr @kaiyuan-cheng just a reminder to have final approvals for this pr.

SamuelTrahanNOAA:
@jkbk2004 There is nothing I can do to effect that end but wait.

bensonr (approved):

jkbk2004:
@kaiyuan-cheng It seems Kaiyuan is not available. @laurenchilutti can you review/merge this pr ? commit queue has been on hold to move to next PR.

kaiyuan-cheng (approved):

jkbk2004:
@laurenchilutti @kaiyuan-cheng @bensonr Thanks for following up with the pr!

PR #224, opened 2022-10-27
Hr ave chem

JianpingHuang-NOAA:
**Description**

This PR is considered as a new feature. It is used to calculate total PM2.5 and its compositions such as NO3, SO4, NH4, EC, OC and their hourly averaged values, as well as hourly-averaged O3, NO2, and NO. The changes are used to support AQM v7 implementation.  The PR requires the change of UPP in order to postprocess mode output files.

**How Has This Been Tested?**
The PR has been only tested on Cactus (WCOSS2).  
Please add the following line in the diag_table as an example to write out hourly-averaged PM2.5.
"gfs_dyn",         "pm25_ave",      "pm25_ave",      "fv3_history",  "all",  .false.,  "none",  2

**Checklist:**

- [X] I have made corresponding changes to the documentation
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/driver/fvGFS/fv_nggps_diag.F90
+++ b/driver/fvGFS/fv_nggps_diag.F90
@@ -73,7 +73,8 @@ module fv_nggps_diags_mod
  use field_manager_mod,  only: MODEL_ATMOS
  use fv_diagnostics_mod, only: range_check, dbzcalc,max_vv,get_vorticity, &
                                max_uh,max_vorticity,bunkers_vector,       &
-                               helicity_relative_CAPS,max_vorticity_hy1
+                               helicity_relative_CAPS,max_vorticity_hy1,  &
+                               average_tracer_hy1,get_pm25
  use fv_arrays_mod,      only: fv_atmos_type
  use module_diag_hailcast, only: do_hailcast, id_hailcast_dhail,         &
                                  id_hailcast_dhail1, id_hailcast_dhail2, &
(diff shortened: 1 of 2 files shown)


bensonr (commented):
Is there a specific reason you are allocating/deallocating the data in fv_diagnostics::get_pm25 instead of allowing it to consume stack memory via explicit declaration?

I'll let @lharris4 decide, but I see no reason for get_pm25 and average_tracer_hy1 to be in fv_diagnostics.F90 if they are only applicable to UFS/NGGPS runs.

lharris4:
Hi, Jianping. Please note that FV3 and the FMS diagnostics manager already supports writing any tracer field to the history files by simply adding the name of the tracer to the diag table. It can also trivially do time-averaging by setting the .false. entry to `ave` (with quotes). For example:
```
"dynamics",  "o3mr",        "o3mr",        "history_file_ave",  "all",  ave,  "none",  2
```
Please note that you need to set a special FMS option to write both instantaneous and time-averaged fields to the same file.

There is no specific PM2.5 output in FV3, but this can be easily done in post; also since FV3 supports many models with many different chemistry packages, which may not define PM2.5 the same way, then this should not be specified within FV3. If you wish to do this on-line, I would strongly suggest defining this diagnostic within the chemistry package.

Thanks,
Lucas

junwang-noaa:
@JianpingHuang-NOAA I suggest you change this PR to draft PR. We can pass pm2.5 from CMAQ as a tracer array to dycore, it is CMAQ's responsibility to compute pm2.5, not dycore.  Raffaele's AQM PR has pm2.5 computed and available in the cap, we just need to pass it back to fv3 to get diagnostic fields for output.

PR #243, opened 2023-02-27
Lightning Threat Indexes: pass w to model

SamuelTrahanNOAA:
**Description**

This is part of a collection of PRs to add lightning threat indexes to the RRFS. The algorithm was originally ported from HRRR to the GFDL_atmos_cubed_sphere NGGPS diagnostics, using the `w` variable in both implementations. I've moved it to CCPP, but it still needs `w` in the formulation, so this three-line change sends `w` to the model.

https://github.com/ufs-community/ufs-weather-model/issues/1632

**How Has This Been Tested?**

A 3km test case from a time that had some lightning was run and compared to the HRRR output. Developers of the scheme found the results to be sufficiently similar. There'll be more extensive testing after this is merged and running in parallels.

**Checklist:**

- [x] I have made corresponding changes to the documentation -- I don't think there are any documentation changes required.
- [x] Any dependent changes have been merged and published in downstream modules -- no downstream changes

### DIFF
--- a/driver/fvGFS/atmosphere.F90
+++ b/driver/fvGFS/atmosphere.F90
@@ -2113,7 +2113,9 @@ contains
          IPD_Data(nb)%Statein%tgrs(ix,k) = _DBL_(_RL_(Atm(mygrid)%pt(i,j,k1)))
          IPD_Data(nb)%Statein%ugrs(ix,k) = _DBL_(_RL_(Atm(mygrid)%ua(i,j,k1)))
          IPD_Data(nb)%Statein%vgrs(ix,k) = _DBL_(_RL_(Atm(mygrid)%va(i,j,k1)))
-         IPD_Data(nb)%Statein%wgrs(ix,k) = _DBL_(_RL_(Atm(mygrid)%w(i,j,k1)))
+         if(associated(IPD_Data(nb)%Statein%wgrs) .and. .not. Atm(mygrid)%flagstruct%hydrostatic) then
+           IPD_Data(nb)%Statein%wgrs(ix,k) = _DBL_(_RL_(Atm(mygrid)%w(i,j,k1)))
+         endif
          IPD_Data(nb)%Statein%vvl(ix,k)  = _DBL_(_RL_(Atm(mygrid)%omga(i,j,k1)))
          IPD_Data(nb)%Statein%prsl(ix,k) = _DBL_(_RL_(Atm(mygrid)%delp(i,j,k1)))   ! Total mass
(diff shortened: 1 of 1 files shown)


junwang-noaa (approved):
The code change is to bring w field from dycore to physics. It looks good to me.

driver/fvGFS/atmosphere.F90
```diff
@@ -2113,6 +2113,7 @@ subroutine atmos_phys_driver_statein (IPD_Data, Atm_block,flip_vc)
          IPD_Data(nb)%Statein%tgrs(ix,k) = _DBL_(_RL_(Atm(mygrid)%pt(i,j,k1)))
          IPD_Data(nb)%Statein%ugrs(ix,k) = _DBL_(_RL_(Atm(mygrid)%ua(i,j,k1)))
          IPD_Data(nb)%Statein%vgrs(ix,k) = _DBL_(_RL_(Atm(mygrid)%va(i,j,k1)))
+         IPD_Data(nb)%Statein%wgrs(ix,k) = _DBL_(_RL_(Atm(mygrid)%w(i,j,k1)))
```
bensonr, line 2116: I believe this should be protected with an if-test for hydrostatic as w is only defined for non-hydrostatic simulations.

lharris4:
When implemented in the physics this check will also need to be made.

driver/fvGFS/atmosphere.F90
SamuelTrahanNOAA: Thank you for noticing that. Also, @junwang-noaa requested that wgrs only be allocated if lightning_threat is enabled, so I'll put both in the same "if" guard.

SamuelTrahanNOAA:
> When implemented in the physics this check will also need to be made.

The code that uses wgrs is never reached unless `lightning_threat .eqv. .true.`

bensonr (approved):
Thanks @SamuelTrahanNOAA.  I haven't looked at the IPD_Data%Statein ddt in a while, but if it's a pointer and not a pure allocatable, then this all looks good.

SamuelTrahanNOAA:
> I haven't looked at the IPD_Data%Statein ddt in a while, but if it's a pointer and not a pure allocatable, then this all looks good

From GFS_typedefs.F90:

```f90
  type GFS_statein_type
    ...
    real (kind=kind_phys), pointer :: wgrs (:,:)   => null()  !< w component of layer wind
    ...
  end type GFS_statein_type

...
  subroutine statein_create (Statein, IM, Model)
    ...
    if(Model%lightning_threat) then
      allocate (Statein%wgrs   (IM,Model%levs))
    endif
```

jkbk2004:
thanks! @bensonr

PR #251, opened 2023-03-23
[dev/emc] Address negative hydrometeors

christinaholtNOAA:
**Description**

Submitted on behalf of @tanyasmirnova

* In fv_sg.F90 set negative graupel to zero in case there are not enough adjacent hydrometeors to compensate for negative values.

* In fv_dynamics.F90 added range_check for negative hydrometeors to identify the location where this happens. No impact on model performance.

The treatment of hydrometeors was problematic for higher-resolution runs and came about while performing experiments with RRFS configurations.

This is a standalone modification that does not have other dependencies.

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Please also note
any relevant details for your test configuration (e.g. compiler, OS).  Include
enough information so someone can reproduce your tests.

For scientific testing, this modification has been included as part of long-term testing with RRFS experimental runs (retros and near real-time), which regularly run on Hera and Jet with Intel compilers. For these RRFS tests, the change has been combined with several other code modifications.

The code has been tested on its own with a recent version of the weather model regression test (e3ad8081 from Mar 14) using a Hera intel configuration. It requires a new baseline and passes all Hera intel tests compared to the new baseline.

**Checklist:**

### DIFF
--- a/model/fv_sg.F90
+++ b/model/fv_sg.F90
@@ -1721,6 +1721,10 @@ contains
              qv2(i,j) = qv2(i,j) - dq
              pt2(i,j) = pt2(i,j) + dq*(icpk(i,j)+lcpk(i,j))
         endif
+!tgs--  Give up and set graupel to zero 
+        if ( qg2(i,j)<0.) then
+             qg2(i,j) = 0.
+        endif
 
 !--------------
 ! Liquid phase:
--- a/model/fv_dynamics.F90
+++ b/model/fv_dynamics.F90
@@ -799,7 +799,19 @@ contains
        if (snowwat > 0) call prt_mxm('snowwat_dyn', q(isd,jsd,1,snowwat), is, ie, js, je, ng, npz, 1.,gridstruct%area_64, domain)
        if (graupel > 0) call prt_mxm('graupel_dyn', q(isd,jsd,1,graupel), is, ie, js, je, ng, npz, 1.,gridstruct%area_64, domain)
        if (hailwat > 0) call prt_mxm('hailwat_dyn', q(isd,jsd,1,hailwat), is, ie, js, je, ng, npz, 1.,gridstruct%area_64, domain)
-     endif
+       if ( flagstruct%range_warn ) then
+       !tgs -  check the location of negative hydrometeors
+        call range_check('liq_wat', q(:,:,:,liq_wat), is, ie, js, je, ng, npz, gridstruct%agrid,   &
+                          -1.e-20, 1.e20, bad_range,fv_time)
+        call range_check('rainwat', q(:,:,:,rainwat), is, ie, js, je, ng, npz, gridstruct%agrid,   &
+                          -1.e-20, 1.e20, bad_range,fv_time)
+        call range_check('graupel', q(:,:,:,graupel), is, ie, js, je, ng, npz, gridstruct%agrid,   &
+                          -1.e-20, 1.e20, bad_range,fv_time)
+        call range_check('snowwat', q(:,:,:,snowwat), is, ie, js, je, ng, npz, gridstruct%agrid,   &
+                          -1.e-20, 1.e20, bad_range,fv_time)
+       endif
+
+     endif ! fv_debug
 #ifdef AVEC_TIMERS
                                                   call avec_timer_stop(6)
 #endif


tanyasmirnova (approved):

model/fv_sg.F90
```diff
@@ -1721,6 +1721,10 @@ subroutine neg_adj3(is, ie, js, je, ng, kbot, hydrostatic, peln, delz, pt, dp,
              qv2(i,j) = qv2(i,j) - dq
              pt2(i,j) = pt2(i,j) + dq*(icpk(i,j)+lcpk(i,j))
         endif
+!tgs--  Give up and set graupel to zero
```
lharris4, line 1724: I would be extremely careful when doing this since this does not conserve water. Perhaps a warning message (with location) should be printed out whenever the adjustment is done, at least if qg2(i,j) < -1.e-20? This would avoid triggering the filling due to -eps values due to machine rounding.

Also, in what circumstances do these negatives appear?

tanyasmirnova: @lharris4 Lucas, I had a communication with you on this a couple months ago. The negative hydrometeors appear after vertical remapping. Call to neg_adj3 fixes most of it, but with Thompson MP that I was testing, negative graupel was still present after neg_adj3. I discovered that there is either no other hydrometeors around these points or the values of other positive hydrometeors around it are not sufficient to cover the negative graupel. Even with positive definite advection option for tracers, this was still happening. The change in the [model/fv_dynamics.F90] adds the diagnostics to find the location of negative hydrometeors.

lharris4:
Hi, Tanya. Thanks for reminding me of our conversation. The cubic-spline
reconstruction is not strictly monotonic and can create undershoots in some
circumstances. We are hoping to re-examine some of the vertical remapping
operations  more closely in the coming months.

Thanks also for adding the additional warning messages, which are very
useful for debugging.

Lucas

tanyasmirnova:
@lharris4 Lucas, thank you for your response and for letting us know about your future plans to resolve the vertical remapping issue.

tanyasmirnova:
@christinaholtNOAA Any progress with merging in this PR?

tanyasmirnova (approved):

tanyasmirnova (approved):

laurenchilutti:
@tanyasmirnova because this is a dev/emc pull request, @christinaholtNOAA or you will need to create a [fv3atm](https://github.com/NOAA-EMC/fv3atm/tree/bcd12a46ff6c581e5e966d1885dc5a28fdc7819d) PR and [ufs-weather-model](https://github.com/ufs-community/ufs-weather-model) PR to get this merged in.

junwang-noaa:
@christinaholtNOAA @tanyasmirnova it looks to me changes you proposed have impact on all the application including global seasonal forecast. Have you tested those applications yet? We are preparing HR2/EP4 experiments, this may change results of those experiments. @yangfanglin would you please ask some dynamics experts to review the code updates?

yangfanglin:
@christinaholtNOAA @tanyasmirnova  This fixing causes a concern for global modeling, especially for forecasts at S2S and climate time scales, which requires mass conservation.   How wide spread are the negative hydrometers caused by remapping ?    Can you share a ppt if you have one ?

yangfanglin:
An experience I can share here is that when we first tested Thompson MP in the global model we found global precipitation keeps increasing.  The cause was traced back to the code where negative tracers were reset to zero before calling the Thompson MP.     For the fixing you proposed, I'd suggest running at least a C96 global forecast for a few month to check its impact on global mean mass and precipitation.

tanyasmirnova:
Similar code is inside mp_thomson.F90:        

 ! Ensure non-negative mass mixing ratios of all water variables
         where(spechum<0) spechum = 1.0E-10     ! COMMENT, gthompsn, spechum should *never* be identically zero.
         where(qc<0)      qc = 0.0
         where(qr<0)      qr = 0.0
         where(qi<0)      qi = 0.0
         where(qs<0)      qs = 0.0
         where(qg<0)      qg = 0.0

yangfanglin:
@tanyasmirnova We first thought the issue of excessive precip in global model was caused by Thompson MP itself but later found the negative hydrometers came from other physics processes including PBL and convection.   We then implemented a positive definite advection scheme (so-called TVD scheme adopted from the Eulerian GSM) in the TKE-EDMF and sa-SAS convection parameterizations.  As for why the mp_thomson.F90 still contains those lines to reset negative tracers,  we should discuss with Greg Thompson to see if it is time to remove them.    Nevertheless, I think it is important to check the impact of code changes on both global and regional models.  ( @gthompsnWRF  @RuiyuSun )

tanyasmirnova:
@yangfanglin Fanglin, we are running a different physics suite in the RRFS regional model, and even with the positive definite advection for tracers we were seeing negative hydrometeors after a vertical remapping. In the global models it is possible that the disputed lines of code will never be used, but they are used in the regional RRFS. I think we should find the way to get the best-possible performance on both ides: global and regional. Maybe we can use the environment variables at the compilation stage?

yangfanglin:
@tanyasmirnova Since this is a change made to the dycore code, I will let @lharris4 comment if it is necessary to safeguard this change (zeroing out negative hydrometers) for global models and if adding a complier flag/directive is an acceptable option.   If you can point me to your branch of the ufs-weather-model with your change,  I can run a low-res global forecast to check the impact on global mass and precipitation.

tanyasmirnova:
@lharris4 @yangfanglin my tests were on hera and jet. 
The codel with this change for graupel is on hera in /scratch1/BMC/gsd-fv3-dev/smirnova/ccpp_fv3/UFS_BaseE_snowcov_24jan23.
I am also wondering if sat_adj=.true. should be used in neg_adj3 with Thompson microphysics. The saturation adjustment is not needed with GFDL microphysics because it is done in there internally, but not in Thompson MP.

RuiyuSun:
The GFDL MP do_sat_adj option can not directly used with Thompson MP. It doesn't have corresponding adjustments to the ice number and rain number. We will need to make a version of adjustment specific to Thompson MP.

tanyasmirnova:
@RuiyuSun @yangfanglin @lharris4 Yes, thank you Ruiyu for your clarification. We are aware that do_sat_adj cannot be used with Thompson MP, but there is local variable **sat_adj** flag inside neg_adj3. I wonder if setting sat_adj=.true. could help with the use of Thompson MP.

lharris4:
Hi, Tanya. This routine probably will not help, because it only works on
cloud and not precipitating condensates. *However* something that may work
is that if negative graupel is found, then if one cannot fill the negative
with graupel from above or below, then fill it instead with cloud ice or
snow from the same grid cell. I would recommend being careful with this
since it effectively introduces an artificial autoconversion to graupel,
but it is probably preferable than the artifical creation of mass at
negative poitns.

I am starting to explore the origins of some of these negatives that we are
also finding in C-SHiELD and were first noticed by our partners at MMM. I
would like to get some more information about these negative graupel
concentrations. Could you point me to a run log with `range_warn = .T.`?

Thanks,
Lucas

tanyasmirnova:
@lharris4 Lucas, we are using hord_tr=10, kord_tr=9 in RRFS.

yangfanglin:
global model is using hord_tr=8 and kord_tr=9

lharris4:
Hi, Tanya, Fanglin. Thanks. @tanyasmirnova could you try hord_tr = 8 in the
case creating negative hydrometeors?

Lucas

tanyasmirnova:
@lharris4 Lucas, I am sorry for the delay with the testing. I was working on this almost a year ago, and the test case has been removed. I will be able to work on this only in May after the EGU.

lharris4:
Hi @tanyasmirnova that is fine; is there anyone else who could easily try this out?

Thanks,
Lucas

tanyasmirnova:
@lharris4 @yangfanglin I ran short forecasts for the recent case  I have (2023030815) with fv_debug=.true. and range_warn=.true. for 4 cases:
1. hord_tr = 10, dt=60, dt_inner = 20, k_split=6, n_split=3, the output is in /scratch2/BMC/wrfruc/smirnova/fv3_rt/RRFS_conus/2023030815/output_hordtr10 directory
2.  hord_tr = 8, dt=60, dt_inner = 20, k_split=6, n_split=3, the output is in /scratch2/BMC/wrfruc/smirnova/fv3_rt/RRFS_conus/2023030815/output_hordtr8 directory
3. hord_tr = 10, dt=36, dt_inner = 36, k_split=2, n_split=5, the output is in /scratch2/BMC/wrfruc/smirnova/fv3_rt/RRFS_conus/2023030815/output_hordtr10_dt36_split25 directory
4. hord_tr = 8, dt=36, dt_inner = 36, k_split=2, n_split=5, the output is in /scratch2/BMC/wrfruc/smirnova/fv3_rt/RRFS_conus/2023030815/output_hordtr8_dt36_split25 directory.
I found that with hord_tr=8 the number of points/levels with negative hydrometeors is significantly reduced, as you can see in slurm* files in each directory.  Case 4 has the least number of points.
Hope you find this helpful for your work.

yangfanglin:
@tanyasmirnova thanks for making these runs and sharing the outcome.  I had also run a few C190L127 global model experiments for 10 to 30 days each, with hord_tr = 8 or hord_tr = 9, and with and without the few lines in Thompson MP (zeroing negative hydrometers) commented out,  I did not see any negative hydrometers.    Although a few negative O3 points in the order of 10E-20 were spotted in some of the runs.    It appears the negative hydrometers you encountered is tied to higher model resolution and stronger vertical velocity.

lharris4:
Hi, Tanya, Fanglin. Thanks for running these tests. The results for hord_tr
10 vs 8 are in-line with the tests I have been running this week. The
hord_tr = 10 scheme is somewhat unreliable when it comes to preserving
monotonicity and especially positivity. We recommend hord_tr = 8 (monotone)
or -5 (positive definite; check code to ensure that this is properly
implemented).

What is "dt_inner"?

Lucas

RuiyuSun:
@lharris4 dt_inner is associated with the inner-loop added in Thompson. For example, dt=60, dt_inner = 20 means Thompson MP is called consecutively three times with the variables updated after each time. 

Ruiyu

bensonr:
There has been no progress on this PR for 9 months, is it still relevant?  If not, can we close it?

bensonr:
With no action/response to the last query (>30 days) this PR is being closed.  Feel free to re-open when/if needed.

PR #259, opened 2023-04-07
add Ubar definition to cases 0, 6, 7, & 9

### LINKED ISSUE #255: Ubar not initialized for some SW dynamics test cases
**Describe the bug**
For at least test_case = 0, 6, 7, 9, there are no initial values given for Ubar.  When running in debug mode, there is a crash in [_init_winds_](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/main/tools/test_cases.F90#L311-#L316) due to undefined Ubar.

**To Reproduce**
Run CI tests for BTwave, BLvortex, or RHwave after compiling solo SW executable.

**Expected behavior**
Expect Ubar to be defined within each select case option in _init_case_

**System Environment**
Describe the system environment, include:
all compilers that do not initialize all variables to zero by default

**Additional context**
N/A

bensonr:
**Description**
This PR fixes undefined values for Ubar when running idealized shallow-water for cases 0, 6, 7, & 9.

Fixes #255 

**How Has This Been Tested?**
It has been tested in idealized debug runs where undefined variables can be trapped via FP exception handling.

**Checklist:**

### DIFF
--- a/tools/test_cases.F90
+++ b/tools/test_cases.F90
@@ -874,6 +874,7 @@
       call mp_stop()
       stop
       case(0)
+         UBar = 0.
          do j=jsd,jed
             do i=isd,ied
 
@@ -1144,6 +1145,7 @@
 
 
       case(6)
+         Ubar = 0.
          gh0  = 8.E3*Grav
          R    = 4.
          omg  = 7.848E-6
@@ -1197,6 +1199,7 @@
 
       case(7)
 ! Barotropically unstable jet
+         Ubar = 0.
          gh0  = 10.E3*Grav
          phis = 0.0
          r0 = radius/12.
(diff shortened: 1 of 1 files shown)


tools/test_cases.F90
```diff
@@ -874,6 +874,7 @@ subroutine init_case(u,v,w,pt,delp,q,phis, ps,pe,peln,pk,pkz,  uc,vc, ua,va, ak,
       call mp_stop()
       stop
       case(0)
+         UBar = 0.
```
lharris4, line 877: Wouldn't it be best to define Ubar = 0. (and the same for Vbar) at the initialization of the variable? Or does that bring up other compiler problems?

laurenchilutti (approved):

PR #283, opened 2023-09-13
Write netcdf axis variables using the same real kind as data variables

DusanJovic-NOAA:
**Description**

Currently, the axis data type in restart files is always "double" regardless of the data type (float vs. double) of used for data variables. This PR updates io routines so that axis data type corresponds to the actual precision used for data variables.
This change is necessary to make restart files bit-identical to the restart files written out by the write grid component in UFS weather model.

Fixes # (issue)
https://github.com/NOAA-EMC/fv3atm/issues/696

**How Has This Been Tested?**

Please describe the tests that you ran to verify your changes. Please also note
any relevant details for your test configuration (e.g. compiler, OS).  Include
enough information so someone can reproduce your tests.

Full regression test of the ufs weather model passed on Hera (both Intel and GNU compilers) and Orion. 

**Checklist:**

Please check all whether they apply or not
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

PR #296, opened 2023-10-16
Cherrypick PR283 from dev/emc to main

laurenchilutti:
**Description**

When cherry-picking the changes from PR #283 I noticed that `model/fv_regional_bc.F90` should also be modified to have an axis type matching the precision that the dycore is built with and not hard-coded double precision.

Fixes # (issue)

**How Has This Been Tested?**

Tested with SHiELD on Gaea

**Checklist:**


lharris4 (approved):
Looks fine, thank you for taking care of this. I see that the calls to `register_field` in atmos_cubed_sphere all use "double". Should we change that too?

laurenchilutti:
Yes I think that we should also change the other register_field calls.  I will make the changes and do some more testing.

bensonr (approved):

PR #297, opened 2023-10-19
Modify external_sst

laurenchilutti:
**Description**

FMS 2023.03 contains mixed precision support and had modified the types of sst_ncep and sst_anom to be hard-coded real8 in amip_interp_mod, which would have required updating some code in this repository when we reference sst_ncep and sst_anom in subroutines (such as pmaxmin).  
In testing this tag of FMS, we noticed that these variables do not need to come from amip_interp_mod as we allocate them within the dycore.  FMS will be updated to no longer declare sst_ncep and sst_anom(see issue: https://github.com/NOAA-GFDL/FMS/issues/1393), and this PR fixes the use statement so that we do not use the amip_interp_mod sst_ncep and sst_anom.  We also removed the `#ifdef NO_GFDL_SHARED` as this is not needed as we will always be compiling the sycore with FMS.

Fixes # (issue)

**How Has This Been Tested?**

Tested with FMS2023.03 with the SHiELD CI and RTS tests

**Checklist:**

### DIFF
--- a/tools/external_sst.F90
+++ b/tools/external_sst.F90
@@ -21,8 +21,9 @@
 
 module external_sst_mod
 
-real, allocatable, dimension(:,:) ::  sst_ncep, sst_anom
 use amip_interp_mod, only: i_sst, j_sst, forecast_mode, use_ncep_sst
+
+real, allocatable, dimension(:,:) ::  sst_ncep, sst_anom
 
 public i_sst, j_sst, sst_ncep, sst_anom, forecast_mode, use_ncep_sst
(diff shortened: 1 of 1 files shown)


bensonr (approved):

lharris4 (approved):
Cleaner is better. Hopefully we can eventually eliminate this routine all together.

bensonr:
Removal depends on all of the nudging routines that use sst_anom and sst_ncep and the two logical triggers from amip_interp

PR #302, opened 2023-10-23
Fix uninitialized values in `cs_profile` when `iv == -3`

### LINKED ISSUE #301: Use of uninitialized values in `fv_mapz_mod.cs_profile` when `iv == -3`
**Describe the bug**

Debugging the source of non-reproducibility for a configuration of SHiELD turned up the use of uninitialized values in a local array in the `cs_profile` subroutine of the `fv_mapz` module.  The issue occurs only when `iv == -3`, which corresponds to one of two possible modes for use when remapping the vertical velocity, the other being `iv == -2`.  I have verified this with additional means, but the issue can be identified by inspecting the lines of code clipped below, considering the case when `iv == -3`:
https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/e94dc4b57c12cc399f6d319dc5733143fc6c718b/model/fv_mapz.F90#L1988-L2071
The issue specifically stems from the local `gam` array, which is declared in line 1988 without any initial assignment.  When `iv == -3` it gets values for level `1` in line 2025 and levels `2` through `km - 1` (inclusive) in line 2034, but is not assigned values for level `km` prior to being used in line 2068.  This leads to undefined / non-reproducible behavior at runtime.

**To Reproduce**

Run SHiELD in non-hydrostatic mode with `fv_core_nml.kord_wz` less than zero and `abs(fv_core_nml.kord_wz) > 7`, e.g. `fv_core_nml.kord_wz = -9`.  This tells the model to use `iv == -3` and `cs_profile` when remapping the vertical velocity:

https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/e94dc4b57c12cc399f6d319dc5733143fc6c718b/model/fv_mapz.F90#L411-L424

**Expected behavior**

I would expect values to be defined for `gam(:,km)` when `iv == -3` prior to it being used in line 2068.

**System Environment**

Describe the system environment, include:
- OS: Ubuntu 20.04
- Compiler(s): GNU version 10.5
- MPI type, and version: MPICH 3.3.2
- netCDF Version: netCDF-C 4.7.3; netCDF-Fortran 4.5.2
- Configure options: 64bit

spencerkclark:
**Description**

Following the suggestion from @lharris4 in https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/301#issuecomment-1775426281, this assigns values to `gam(i,km)` consistent with the interior of the column when `iv == -3` in `cs_profile`.  It should be expected to change answers for non-hydrostatic simulations where `fv_core_nml.kord_wz < -7`. 

Fixes #301

**How Has This Been Tested?**

This has not been tested for scientific correctness, though I have tested that the change proposed here addresses the non-reproducibility issue identified in #301.  I now get bitwise-identical results every time I run the model in this configuration.

**Checklist:**

### DIFF
--- a/model/fv_mapz.F90
+++ b/model/fv_mapz.F90
@@ -2044,6 +2044,7 @@ else ! all others
        grat = delp(i,km-1) / delp(i,km)
        q(i,km) = ( 3.*(a4(1,i,k-1)+d4(i)*a4(1,i,k)) - grat*qs(i) - q(i,k-1) )/bet
        q(i,km+1) = qs(i)
+       gam(i,km) = d4(i) / bet
     enddo
 
   else ! all others


lharris4 (approved):
This should be correct. Thank you for submitting a fix so quickly.

@laurenchilutti can we bring this back into GitLab? Thanks.

spencerkclark:
@lharris4 thanks for quickly providing the suggestion for the fix!  

FYI my force push was only to update the commit message to indicate all values of `kord_wz` for which this is relevant, instead of referring just to `kord_wz = -9`.

spencerkclark:
> FYI my force push was only to update the commit message to indicate all values of kord_wz for which this is relevant, instead of referring just to kord_wz = -9.

Apologies for potentially interfering with the CI with this rapid re-push.

bensonr (approved):

lharris4:
@spencerkclark No problem, thank you for taking care of this.

PR #316, opened 2024-01-24
Port `delp_dycore` diagnostic to public version of SHiELD

spencerkclark:
**Description**

This PR adds a diagnostic for the pressure thickness which includes the mass of water condensates (`delp_dycore`).  This is the actual pressure thickness used internally by the dynamical core to define things like tracer mixing ratios, and so it is useful to be able to output it for diagnosing budgets.  In many ways it is a simpler diagnostic to implement than the existing `delp` diagnostic.

xref: I implemented this diagnostic as part of another merge request to the internal version of SHiELD (see [this commit](https://gitlab.gfdl.noaa.gov/fv3team/atmos_cubed_sphere/-/merge_requests/134/diffs?commit_id=803a79cc4e8b8b0e5c175790af88282fff32d008); note authentication is needed to access the link).  

Since this is such a simple change, and one that already exists in the internal version of SHiELD, I hope we can review / merge it quickly, as I need it for a new simulation.  Thanks!

cc: @lharris4 and @bensonr

**How Has This Been Tested?**

I tested this in a short C12 resolution simulation within a Docker container, and validated that I could reproduce the existing `delp` diagnostic using this new `delp_dycore` diagnostic and the sum of the condensate tracers ([notebook](https://gist.github.com/spencerkclark/aa83bdecf4e6a5dce31692b826d3a104)).

**Checklist:**

### DIFF
--- a/tools/fv_diagnostics.F90
+++ b/tools/fv_diagnostics.F90
@@ -960,8 +960,10 @@ contains
           !            'Relative Humidity', '%', missing_value=missing_value, range=rhrange )
           id_delp = register_diag_field ( trim(field), 'delp', axes(1:3), Time,        &
                'pressure thickness'//massdef_str, 'pa', missing_value=missing_value )
-          id_delp_dycore = register_diag_field ( trim(field), 'delp_dycore', axes(1:3), Time,        &
-               'pressure thickness as seen by the dynamical core', 'pa', missing_value=missing_value )
+#ifdef GFS_PHYS
+          id_delp_total = register_diag_field ( trim(field), 'delp_total', axes(1:3), Time,        &
+               'FV3 pressure thickness (dry air + all water species)', 'pa', missing_value=missing_value )
+#endif
           if ( .not. Atm(n)%flagstruct%hydrostatic )                                        &
                id_delz = register_diag_field ( trim(field), 'delz', axes(1:3), Time,        &
                'height thickness', 'm', missing_value=missing_value )
@@ -3014,8 +3016,6 @@ contains
           endif
        endif
 
-       if(id_delp_dycore > 0) used=send_data(id_delp_dycore, Atm(n)%delp(isc:iec,jsc:jec,:), Time)
-
 #ifdef GFS_PHYS
        if(id_delp > 0 .or. id_cape > 0 .or. id_cin > 0 .or. &
             ((.not. Atm(n)%flagstruct%hydrostatic) .and. (id_pfnh > 0 .or. id_ppnh > 0)) .or. &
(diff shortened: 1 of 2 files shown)


lharris4 (approved):
This is good, thank you. Note that if `GFS_PHYS` is not defined then `delp_dycore` and `delp` are identical: the version used when `GFS_PHYS` is defined is to match the moist mass (no condensates) used in the Legacy GFS and expected by downstream applications of the GFS.

tools/fv_diagnostics.F90
```diff
@@ -3012,6 +3014,7 @@ subroutine fv_diag(Atm, zvir, Time, print_freq)
           endif
        endif
 
+       if(id_delp_dycore > 0) used=send_data(id_delp_dycore, Atm(n)%delp(isc:iec,jsc:jec,:), Time)
```
bensonr, line 3017: @lharris4 - should this be moved inside the #ifdef to not have the same diagnostic output with two different names when the macro GFS_PHYS is not defined?

lharris4: I think you are right, there is no need to carry two definitions in non-`GFS_PHYS` implementations. 

Ideally `delp` or `delp_total` would be the regular total-mass definition (dependent on nwat), `delp_moist` or `delp_dry` defined appropriately, for all configurations, and the NetCDF metadata would make clear which is which. 

`delp` is typically not useful for most end-users, only to those creating diagnostics very closely consistent with the numerics, so this would not be too disruptive.

spencerkclark: > should this be moved inside the #ifdef to not have the same diagnostic output with two different names when the macro GFS_PHYS is not defined?

@bensonr indeed in the current state I think this makes sense.  

> Ideally `delp` or `delp_total` would be the regular total-mass definition (dependent on nwat), `delp_moist` or `delp_dry` defined appropriately, for all configurations, and the NetCDF metadata would make clear which is which.
> 
> `delp` is typically not useful for most end-users, only to those creating diagnostics very closely consistent with the numerics, so this would not be too disruptive.

@lharris4 do you have a concrete suggestion?  I am open to making a breaking change—the mismatch between the `delp` diagnostic and the `delp` variable in restart files caught me off guard originally.  I'm also fine with changing the name of the diagnostic added here, e.g. `delp_total` is what I named it in FV3GFS, and maybe improving the metadata of the diagnostics for the two versions of the pressure thickness.

spencerkclark: I went ahead with a conservative approach—renaming `delp_dycore` to `delp_total`, and only exposing it as a diagnostic when compiled using `GFS_PHYS`.  I also revised the description to hopefully be a little more explicit:

```
'FV3 pressure thickness (dry air + all water species)'
```  

I notice the description of `delp` in SHiELD does at least include a reference to `'GFS moist-mass'` if compiled with `GFS_PHYS` ([which is better than FV3GFS](https://github.com/ai2cm/fv3gfs-fortran/blob/e3c522d285b2c9dac925f21e8c3d1f36df9f2dda/FV3/atmos_cubed_sphere/tools/fv_diagnostics.F90#L747-L748)), so I have left that alone.

@lharris4 @bensonr let me know what you think.  Thanks!

bensonr (approved):

lharris4 (approved):
Simple is good. Thanks for making these changes.

PR #274, opened 2023-06-07
Add capability to read IAU increments in cubed sphere format

mark-a-potts:
This PR adds the capability to read in NetCDF-based cubed-sphere increments generated by JEDI. It includes a new (optional) namelist option to specify that non-gaussian formatted increment files are to be read. 

Fixes # (issue)

**How Has This Been Tested?**

Two unit tests have been added to the code and can be run via ctest after the build. 

**Checklist:**

Please check all whether they apply or not
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

Issue #188, opened 2022-04-29
Read cube sphere grid increments for DA/IAU


CoryMartin-NOAA:
**Is your feature request related to a problem? Please describe.**
In anticipation of transitioning from GSI to JEDI for UFS DA capabilities, we would like FV3 to be able to read increments on the native cube sphere grid in addition to the Gaussian/Lat Lon grid and then interpolated like is currently supported. 

**Describe the solution you'd like**
Now that the UFS-weather-model write-grid component can write out history files on the native grid (with variables with dimensions such as:  `t(tile, layer, lon, lat)`), and FV3-JEDI can read and write these files, being able to read a similarly configured/formatted file for analysis increments would be preferred. Additionally, a configuration table to map model variables to increment file field names would ensure maximum compatibility across different applications.

**Describe alternatives you've considered**
The exact format of the files can still be up for discussion provided that it is flexible for dynamics, tracers, and possibly surface variables, and preferably contains every tile as a dimension in one file.

**Additional context**
EMC/PSL intend to perform the majority of this work but would like to confirm approach/design with GFDL before proceeding.

CoryMartin-NOAA:
Cannot assign people, but tagging @jswhit2 @pjpegion @catherinethomas-noaa @russtreadon-noaa for awareness

bensonr:
@CoryMartin-NOAA - for seamless interoperability with existing FMS io infrastucture and performance, it is best the files be constructed similar to a restart file.  Instead of a single file with data for all of the tile information, there would be a file specific to each tile.  This makes it easier to account for the different rotations as well as handling nested configurations.

lharris4:
Hi, all. Earlier GMAO was doing native-FV3 grid DA on the same lines. Has
there been any coordination with them?

Thanks,
Lucas

CoryMartin-NOAA:
I am not sure if @danholdaway plans to read increments for GEOS or not, but he and I have been coordinating on the history file IO to be compatible in FV3-JEDI between UFS and GEOS.

@bensonr would you suggest something like `20220429.120000.fv_increment.tile1.nc` in the FMS RESTART format? I think I'm fine with that provided that it is just one increment per tile per IAU time. I'm hesitant to have something like `20220429.120000.fv_tracer.tile1.increment.nc` because of the number of files that would entail.

danholdaway:
[...]

PR #342, opened 2024-05-17
Feature/cubed sphere inc

DavidNew-NOAA:
**Description**

This PR adds the ability to read increments on the native cubed sphere grid, both inside and outside the IAU framework. This option can be turned on by an input namelist option, "increment_file_on_native_grid", which defaults to .false.. 

The motivation for this PR is that we would like to read cubed sphere increments in the Global Workflow used by NOAA-EMC for the new JEDI DA system.

This PR is based off of another PR, #274, but I'm opening a new one, because I had to make so many changes to allow increments outside of the IAU framework, since IAU is not yet implemented in JEDI.

Overview of code changes:

A new module, "cubed_sphere_inc_mod.F90", has been added to the "tools" directory, and it has a subroutine, "read_cubed_sphere_inc", that will read the increment into an "increment_data_type" structure. This structure was formerly the "iau_internal_data_type" data structure in "fv_iau_mod.F90", but I've moved it into the new module to avoid circular dependencies and renamed it to reflect the fact that it's not just for IAU.

"read_cubed_sphere_inc" is called in either "fv_iau_mod.F90" for "fv_treat_da_inc.F90", depending on whether or not IAU is used. If not using IAU, a new subroutine, "read_da_inc_cubed_sphere", in "fv_treat_da_inc.F90" is called in "fv_restart.F90" to read and apply the increment. This involves some halo updates and a physics call so the A-grid wind increment can be applied to the D-grid wind.

Fixes # (issue)

N/A

**How Has This Been Tested?**

I ran a few of sets of experiments that involved three C96, 6-hour forecasts with identical initial conditions. These experiments were run on RDHPCS Hera.

First, I used JEDI to produce two increments, identical except that one was on the native grid and one was interpolated to the Gaussian grid, and then I ran two forecasts, one for each increment (not with IAU). In the third experiment, I ran the forecast without an increment. I then subtracted the free (no increment) forecast model fields from the model fields of the forecasts using the increments. This allows me to isolate the impact of the increments on the forecast. Finally, I plotted the resulting model fields and their differences and verified visually that the impact of the increments were identical in terms of large-scale patterns. Naturally there are differences between the two forecasts because of interpolation, but these appeared to produce mostly noise. But the fact that there are (expected) differences is why I had to verify my results visually.

The following is a plot of "o3mr" at model level 20 (statosphere), using OMI satellite observations for the increment:

Figure_1

Here is another plot of "ua" at model level 20, using satwind observations for the increment:

Figure_2

**Checklist:**

### DIFF
--- a/tools/cubed_sphere_inc_mod.F90
+++ b/tools/cubed_sphere_inc_mod.F90
@@ -45,7 +45,7 @@ module cubed_sphere_inc_mod
       ! Read increments
       call get_var5_r4( ncid, 'u_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%ua_inc )
       call get_var5_r4( ncid, 'v_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%va_inc )
-      call get_var5_r4( ncid, 'temp_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%temp_inc )
+      call get_var5_r4( ncid, 'T_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%temp_inc )
       call get_var5_r4( ncid, 'delp_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%delp_inc )
       if ( .not. Atm%flagstruct%hydrostatic ) then
          call get_var5_r4( ncid, 'delz_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%delz_inc )
--- a/tools/cubed_sphere_inc_mod.F90
+++ b/tools/cubed_sphere_inc_mod.F90
@@ -1,0 +1,69 @@
+module cubed_sphere_inc_mod
+
+  use tracer_manager_mod, only: get_tracer_index, get_number_tracers, get_tracer_names
+  use field_manager_mod,  only: MODEL_ATMOS
+  use fv_arrays_mod,      only: fv_atmos_type
+  use sim_nc_mod,         only: open_ncfile, get_var5_r4, check_var_exists, close_ncfile
+
+  implicit none
+  type increment_data_type
+    real, allocatable :: ua_inc(:,:,:)
+    real, allocatable :: va_inc(:,:,:)
+    real, allocatable :: temp_inc(:,:,:)
+    real, allocatable :: delp_inc(:,:,:)
+    real, allocatable :: delz_inc(:,:,:)
+    real, allocatable :: tracer_inc(:,:,:,:)
+  end type increment_data_type
+
+  public read_cubed_sphere_inc, increment_data_type
+
+  contains
+
+!----------------------------------------------------------------------------------------
+
+    subroutine read_cubed_sphere_inc(filename, increment_data, Atm)
+      character(*), intent(in)                 :: filename
+      type(increment_data_type), intent(inout) :: increment_data
+      type(fv_atmos_type), intent(in)          :: Atm
+
+      integer           :: isc, iec, jsc, jec, npz, itracer, ntracers, itile
+      character(len=64) :: tracer_name
+      integer           :: ncid, ncerr
+      
+      ! Get various dimensions
+      call get_number_tracers(MODEL_ATMOS, num_tracers=ntracers)
+      isc = Atm%bd%isc
+      iec = Atm%bd%iec
+      jsc = Atm%bd%jsc
+      jec = Atm%bd%jec
+      npz = Atm%npz
+      itile = Atm%tile_of_mosaic
+
+      ! Open increment file
+      call open_ncfile( trim(filename), ncid )
+
+      ! Read increments
+      call get_var5_r4( ncid, 'u_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%ua_inc )
+      call get_var5_r4( ncid, 'v_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%va_inc )
+      call get_var5_r4( ncid, 'temp_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%temp_inc )
+      call get_var5_r4( ncid, 'delp_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%delp_inc )
+      if ( .not. Atm%flagstruct%hydrostatic ) then
+         call get_var5_r4( ncid, 'delz_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%delz_inc )
+      end if
+
+      ! Read tracer increments
+      do itracer = 1,ntracers
+         call get_tracer_names(MODEL_ATMOS, itracer, tracer_name)
+         call check_var_exists(ncid, trim(tracer_name)//'_inc', ncerr)
+         if ( ncerr == 0 ) then
+            call get_var5_r4( ncid, trim(tracer_name)//'_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%tracer_inc(:,:,:,itracer) )
+         end if
+      end do
+         
+      ! Close increment file
+      call close_ncfile(ncid)
+    
+  end subroutine read_cubed_sphere_inc
+
+!----------------------------------------------------------------------------------------
+end module cubed_sphere_inc_mod
--- a/tools/fv_treat_da_inc.F90
+++ b/tools/fv_treat_da_inc.F90
@@ -159,7 +159,7 @@ module fv_treat_da_inc_mod
 contains
   !=============================================================================
   !>@brief The subroutine 'read_da_inc' reads the increments of the diagnostic variables
-  !! from the DA-generated files.
+  !! from the DA-generated files on a lot-lon grid.
   !>@details Additional support of prognostic variables such as tracers can be assessed
   !! and added upon request.
   !>@author Xi.Chen <xi.chen@noaa.gov>
@@ -472,6 +472,11 @@ contains
     !---------------------------------------------------------------------------
   end subroutine read_da_inc
 
+  !=============================================================================
+  !>@brief The subroutine 'read_da_inc_cubed_sphere' reads increments of diagnostic variables
+  !! from DA-generated files on the native cubed-sphere grid.
+  !>@author David.New <david.new@noaa.gov>     
+  !>@date 05/16/2024  
   subroutine read_da_inc_cubed_sphere(Atm, fv_domain, bd, npz_in, nq, &
                                       u, v, q, delp, pt, delz, &
                                       is_in, js_in, ie_in, je_in, isc_in, jsc_in, iec_in, jec_in)
--- a/tools/fv_treat_da_inc.F90
+++ b/tools/fv_treat_da_inc.F90
@@ -468,7 +468,8 @@ contains
     !---------------------------------------------------------------------------
   end subroutine read_da_inc
 
-  subroutine read_da_inc_cubed_sphere(Atm, fv_domain, bd, npz_in, nq, u, v, q, delp, pt, delz, &
+  subroutine read_da_inc_cubed_sphere(Atm, fv_domain, bd, npz_in, nq, &
+                                      u, v, q, delp, pt, delz, &
                                       is_in, js_in, ie_in, je_in, isc_in, jsc_in, iec_in, jec_in)
     type(fv_atmos_type),       intent(in)    :: Atm
     type(domain2d),            intent(inout) :: fv_domain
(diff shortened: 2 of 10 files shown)


tools/fv_treat_da_inc.F90
```diff
@@ -140,15 +149,17 @@ module fv_treat_da_inc_mod
...
   private
 
-  public :: read_da_inc,remap_coef
+  public :: read_da_inc, read_da_inc_cubed_sphere, remap_coef
 
 contains
   !=============================================================================
   !>@brief The subroutine 'read_da_inc' reads the increments of the diagnostic variables
-  !! from the DA-generated files.
+  !! from the DA-generated files on a lot-lon grid.
```
lharris4, line 162: "lat-lon"

tools/fv_treat_da_inc.F90
```diff
@@ -461,6 +472,183 @@ end subroutine apply_inc_on_3d_scalar
...
+#else
+    o3mr    = get_tracer_index(MODEL_ATMOS, 'o3mr')
+#endif
+
+    ! Apply increments
+    do k = 1,npz_in
+       do j = jsc_in,jec_in
+          do i = isc_in,iec_in
+             pt(i,j,k)   = pt(i,j,k)   + increment_data%temp_inc(i,j,k)
+             delp(i,j,k) = delp(i,j,k) + increment_data%delp_inc(i,j,k)
```
lharris4, line 633: Do make sure that the delp increment is consistent with the tracer increments, as tracers are defined as the total-mass specific ratio and delp the mass of dry air, water vapor, and all condensates.

lharris4 (commented):
I think this is good. One thing I do recommend is checking for the consistency of the increments for both delp and the tracers; this should be OK as long as JEDI doesn't do any transformations on these variables.

tools/fv_treat_da_inc.F90
DavidNew-NOAA: @lharris4 You mean in JEDI, right? That's a good point, because while I have to look at how delp is defined in JEDI, I do know that the delz increment is currently defined hydrostatically as the difference between the hydrostatic delz and the beginning and end of the cycle, and that hydrostatic relationship is determined using a virtual temperature that only accounts for dry air and water vapor, but no condensates. It's something I've been meaning to address.

lharris4: @DavidNew here I am referring to `delp` and the adjustment of tracer masses when `delp` is incremented, as is done in `fv_update_phys()`. This is because each tracer is defined as the specific ratio of tracer mass and total air mass (`delp`) in a grid cell, and when `delp` is updated the denominator is changed; so to ensure tracer mass is conserved the tracers need to be adjusted to account for the change in `delp`

DavidNew-NOAA: Got it. I'll make a note to look into verifying this in JEDI. Thanks

DavidNew-NOAA:
Are there any other obstacles to getting this PR reviewed and merged?

lharris4 (approved):
I think this is good. Thanks for submitting this.

tools/cubed_sphere_inc_mod.F90
```diff
@@ -0,0 +1,69 @@
...
+      do itracer = 1,ntracers
+         call get_tracer_names(MODEL_ATMOS, itracer, tracer_name)
+         call check_var_exists(ncid, trim(tracer_name)//'_inc', ncerr)
+         if ( ncerr == 0 ) then
+            call get_var5_r4( ncid, trim(tracer_name)//'_inc', isc,iec, jsc,jec, 1,npz, itile, 1, increment_data%tracer_inc(:,:,:,itracer) )
+         end if
+      end do
+         
+      ! Close increment file
+      call close_ncfile(ncid)
```
bensonr, line 64: You may get a complaint from operations that this is too disruptive to the filesystem, especially for large numbers of mpi-ranks.  You have every core opening the file, performing reads and then closing the file.  While I know you are using these embedded I/O routines from the dycore, they were originally put in as hacks a long time ago and generally won't play nice now that we've scaled up to thousands or cores on distributed filesystems.

You'd be better off using the fms2_io routines which are already domain aware (suited for native grid) and will read in the correct increment for a given mpi-rank.  The fms2_io routines have recently been modified by UFS contributors to use MPI-IO for reads of NetCDF4 files ([FMS PR#1477](https://github.com/NOAA-GFDL/FMS/pull/1477)).

DavidNew-NOAA:
@bensonr See one of my original 
> Note: There was a suggestion in the comments of #274 (cc lharris4 bensonr) to use FMS calls for NetCDF IO, but I was unable to get that to work. I tried using the "read_data" FMS2 subroutine to read the increments using "corner" and "edge_lengths" arguments for start and count, but for some reason it would always read the first tile even if I specified the nth tile. However, when I fed the "corner" and "edge_lengths" arguments directly into the "nf90_get_var" subroutine from the NetCDF library, this problem disappeared. This makes me think there may be a bug in the 5D "read_data" subroutine, unless I'm missing something. In the end I decided to extend the "sim_nc_mod.F90" module with a new subroutine since up until now, that module has been used to read increments, not FMS.

I would really like to use fms2_io, but in addition to the above problem, I could not find an example to work from where someone does a 5D read from a cubed-sphere history file, let alone using domain awareness features. Every example I found was for reading FMS restart files.  

Moreover, as of right now, Gaussian grid increment files are read, both IAU and non-IAU, using the sim_nc_mod.F90 module routines anyway.

DavidNew-NOAA:
> @bensonr See one of my original
> 
> > Note: There was a suggestion in the comments of #274 (cc lharris4 bensonr) to use FMS calls for NetCDF IO, but I was unable to get that to work. I tried using the "read_data" FMS2 subroutine to read the increments using "corner" and "edge_lengths" arguments for start and count, but for some reason it would always read the first tile even if I specified the nth tile. However, when I fed the "corner" and "edge_lengths" arguments directly into the "nf90_get_var" subroutine from the NetCDF library, this problem disappeared. This makes me think there may be a bug in the 5D "read_data" subroutine, unless I'm missing something. In the end I decided to extend the "sim_nc_mod.F90" module with a new subroutine since up until now, that module has been used to read increments, not FMS.
> 
> I would really like to use fms2_io, but in addition to the above problem, I could not find an example to work from where someone does a 5D read from a cubed-sphere history file, let alone using domain awareness features. Every example I found was for reading FMS restart files.
> 
> Moreover, as of right now, Gaussian grid increment files are read, both IAU and non-IAU, using the sim_nc_mod.F90 module routines anyway.

@bensonr Are you aware of an example of such a 5D read with FMS2 using domain awareness capability?

lharris4:
@DavidNew-NOAA I am curious: what is the use case for a 5D I/O statement? Usually, 4d data (x,y,z, tracer_index) is sufficient, and the time dimension is read one slice at a time.

Thanks
Lucas

DavidNew-NOAA:
@lharris4 Increment files in cubed-sphere history format have dimensions (x,y,x,tile,time). The time dimension is always length=1, but it exists. Tiles are read one at a time.

DavidNew-NOAA:
@bensonr @lharris4 Is there someone in the FMS community that would know how to implement a domain aware 5D read of a cubed sphere history file?

laurenchilutti:
@DavidNew-NOAA Lucas and Rusty are both out of the office today, but I can point you to an example of a 5D read from FMS: https://github.com/NOAA-GFDL/FMS/blob/main/test_fms/fms2_io/test_domain_io.F90#L125-L136

DavidNew-NOAA:
Awesome, thank you @laurenchilutti

DavidNew-NOAA:
@laurenchilutti What I don't understand here https://github.com/NOAA-GFDL/FMS/blob/e81d5b171861e8f3723c2648bc5b76f50bda94a1/test_fms/fms2_io/test_domain_io.F90#L126 is why the tile dimension does not need to be registered. How does FMS know which dimension is the tile dimension?

laurenchilutti:
The tile dimension does not need to be registered here because you are reading the file, and FMS will get the details on the tile axis from reading the file.  You still need to register the x and y dimension because the file is domain decomposed.

laurenchilutti:
here is a section of the fms2_io readme with detail on this topic: https://github.com/NOAA-GFDL/FMS/blob/main/fms2_io/readme.md#d-readingwritting-non-restarts

DavidNew-NOAA:
@laurenchilutti Thanks for the info and resources. Will give this a try

DavidNew-NOAA:
@laurenchilutti So I did a bit of a deep dive on FMS2 IO, and I’m fairly certain now about the inability of the FMS domain-decomposed read subroutines to infer from a NetCDF file dimension is the tile dimension when it reads the file.  It seems like there would need to be a way to register the ```tile``` dimension just like the ```x``` and ```y``` dimensions are registered. Then when the  ```is_variable_domain_decomposed``` subroutine is called [here]( https://github.com/NOAA-GFDL/FMS/blob/eeedbab8ece4d68ff81d2cebcc350f659cc12982/fms2_io/include/domain_read.inc#L714) in the ```domain_read_5d``` subroutine, you could obtain something like a ```tiledim_index``` just like there’s a ```xdim_index``` and ```ydim_index``` that you get when you register the ```x``` and ```y``` dimensions. Then you can set the ```corner``` parameter for the ```netcdf_read_data``` call [here]( https://github.com/NOAA-GFDL/FMS/blob/eeedbab8ece4d68ff81d2cebcc350f659cc12982/fms2_io/include/domain_read.inc#L822) so that it accesses the appropriate tile, similar to how that parameter is set for the ```x``` and ```y``` dimensions [here]( https://github.com/NOAA-GFDL/FMS/blob/eeedbab8ece4d68ff81d2cebcc350f659cc12982/fms2_io/include/domain_read.inc#L740).

Now, as I mentioned in the description of this PR, I tried using FMS using what I thought was an unstructured domain read by passing optional ```corner``` and ```edge_lengths``` parameters to the ```read_data``` subroutine, but I couldn’t understand why regardless of which tile I tried to read, I would always read the first tile. It turns out that unless you pass an optional ```broadcast``` parameter, the ```read_data``` wrapper uses the ```domain_read_5d``` subroutine and not the ```unstructured_domain_read_5d``` subroutine. But, even though I am passing  ```corner``` and ```edge_lengths```, when ```is_variable_domain_decomposed``` returns ```.true.``` [here]( https://github.com/NOAA-GFDL/FMS/blob/eeedbab8ece4d68ff81d2cebcc350f659cc12982/fms2_io/include/domain_read.inc#L714C13-L714C42), those parameters are ignored, and the ```corner``` parameter defaults to ```1``` [here]( https://github.com/NOAA-GFDL/FMS/blob/eeedbab8ece4d68ff81d2cebcc350f659cc12982/fms2_io/include/domain_read.inc#L725). Hence, I always read the first tile regardless of which tile I specify. This isn’t really a bug, but it is a very opaque feature that’s hard to understand without looking at the code. It would be better if it just errored when you specify ```corner``` and ```edge_lengths``` but ```is_variable_domain_decomposed``` returns ```.true.```.

So, there’s two ways forward. I can go the easy way of making sure I’m calling ```unstructured_domain_read_5d``` by passing the ```broadcast``` parameter to the ```read_data``` wrapper, but bypassing the nice domain awareness features that FMS has. Alternatively, I could help implement the option in FMS to register the ```tile``` dimension like I discussed above. That would be more work, but it would also be a really a nice feature, which, from all the discussion already in this PR, the community expect to already be there. In a sense, the ```tile``` dimension IS domain decomposed, because each processor corresponds to exactly one tile in FV3. I think you are an FMS contributor, no? Would this be something useful for me to add to FMS?

bensonr:
@DavidNew-NOAA - it might be time for a meeting to ensure we all have a full understanding of the cubed-sphere and how it is represented in FMS for domains, domain-aware I/O, etc.  If you'd prefer I write it all down here, I can.  I'd also like to understand why the IAU increment file includes all tiles instead of partitioning into individual files for each tile.  I think once we have a discussion, you'll understand our choices related to I/O within FMS.

DavidNew-NOAA:
@bensonr I think a meeting is a good idea and would be more efficient.

DavidNew-NOAA:
@laurenchilutti @lharris4 @bensonr @XiaqiongZhou-NOAA I made changes to this PR, now, as requested, using FMS calls to read the increment files. The increment files are now split into one file for each tile. When the namelist parameters ```read_increment``` and ```increment_file_on_native_grid``` are both set to ```.true.```, ```res_latlon_dynamics``` becomes the prefix for the increment file names, rather than the full name itself. Thus, for example, if ```res_latlon_dynamics='atminc'```, the increment files will be named ```atminc.tile1.nc```, ```atminc.tile2.nc```, etc. 

Can you please perform review of these changes? Thanks

lharris4 (approved):

bensonr (approved):
@DavidNew-NOAA - thanks for being so accommodating with the suggestions to use the existing IO layers within FMS.

XiaqiongZhou-NOAA (approved):

laurenchilutti (approved):

bensonr:
According to the comments, may address issue #188 (close when merged).

DavidNew-NOAA:
@laurenchilutti Is there anything I need to do to move this PR along now that everything is approved?

laurenchilutti:
Once approvals are done, testing needs to be done on the UFS PR associated with this change.  And then someone usually tells me to merge when the UFS PR is ready to merge.  @jkbk2004 is there anything @DavidNew-NOAA needs to do to get this merged?

DavidNew-NOAA:
If I'm not mistaken there following error in testing is due to https://github.com/DavidNew-NOAA/GFDL_atmos_cubed_sphere/commit/7c3102fccc493d4c006c936ea88e4b512f6f5251 :

https://github.com/NOAA-EMC/fv3atm/actions/runs/10304307448/job/28522464673?pr=837#step:6:2384

Does anyone have some insight on this?

laurenchilutti:
The commit you reference was just merged in today and is accompanied by an FV3atm PR that was also merged today: https://github.com/NOAA-EMC/fv3atm/pull/798. I would make sure that your FV3atm PR has merged in the latest from the develop branch.

jkbk2004:
@DavidNew-NOAA We need to update the namelist file option (increment_file_on_native_grid) at weather model regression test cases. Can you create a pull request to https://github.com/ufs-community/ufs-weather-model ?

jkbk2004:
I missed. @DavidNew-NOAA has a draft pr https://github.com/ufs-community/ufs-weather-model/pull/2304. I think we can make the PR ready with namelist parameter update and pre-test.

DavidNew-NOAA:
@jkbk2004 I added the namelist parameter in https://github.com/NOAA-EMC/fv3atm/pull/837 for FV3. Where do I need to modify UFS? I was able to test successfully with the namelist parameter in FV3

zach1221:
@bensonr testing is complete on UFS-WM PR #2389. Can you please merge this PR for us?

tools/fv_treat_da_inc.F90
danholdaway: @lharris4 sorry to take you back a long time, we have just been discussing something at EMC. Is the process you're referring to done using the remap methods normally? I wonder if this is similar to what is needed when starting the model from a cold start. Is it true to say (in some sense) that once an increment is added to the state of the model you have a 'cooler start' because the incrementing of pressure moves all the model layers around?

lharris4: Hi, Dan. We perform the adjustments when doing the intialization, as in `external_ic_mod::nggps_ic()`. I would presume that if you are doing DA cycling no adjustments are necessary because the cycling is done on outputs from the same driving model. I am simply making a cautionary note to make sure that the DA cycling system isn't transforming the increments into something unexpected. One possibility is that if inconsistent increments are applied to both delp and tracers then there may be an unphysical change in the dry-air mass.

danholdaway: Thanks for the response Lucas. That's a great point about the need to balance how the tracers are defined. Are there routines for doing that in FV3?

Is it not also true that the levels of the model are changed if DELP is changed? And thus that the variables stored on those levels, temperature, winds etc also need to be interpolated vertically to the new levels? Perhaps there is a difference between when running hydrostatically and non-hydrostatically here to consider as well.

lharris4: Hi, Dan. This is done within this block of code here:

[https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/main/model/fv_update_phys.F90#L328-L367](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/blob/main/model/fv_update_phys.F90#L328-L367)

For your purposes, `adj_mass_vmr = 0`.

Please let me know if you need any further help.

Thanks
Lucas

PR #357, opened 2024-09-09
Minor corrections for mixed mode support in AM5

uramirez8707:
**Description**
This PR includes some fixes needed for mixed mode support in AM5:

1. Return the correct version of the `lon` and `lat` in `atmosphere_boundary`
2. Fix a typo so that the correct constants in `fv_grid_tools.F90` are included
3. Added missing parenthesis when setting `z_bot` in `get_bottom_mass`, this was causing answer changes when using the non-mixed precision version of am5

Fixes # (issue)

**How Has This Been Tested?**
Ran AM5 with and without mixed precision

**Checklist:**
Please check all whether they apply or not
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published in downstream modules

PR #360, opened 2024-10-01
allows mix precision compilation

JosephMouallem:
Allows a mixed precision compilation of the dycore. FV3 can be compiled in 32bit and works with other components compiled in 64bit. This depends on:
https://github.com/NOAA-GFDL/SHiELD_physics/pull/56
https://github.com/NOAA-GFDL/SHiELD_build/pull/46

Inspired from https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/357

### DIFF
--- a/driver/SHiELDFULL/atmosphere.F90
+++ b/driver/SHiELDFULL/atmosphere.F90
@@ -31,8 +31,13 @@ module atmosphere_mod
 !-----------------
 ! FMS modules:
 !-----------------
+use platform_mod, only: r8_kind, r4_kind
 use block_control_mod,      only: block_control_type
-use constants_mod,          only: cp_air, rdgas, grav, rvgas, kappa, pstd_mks, pi
+#ifdef OVERLOAD_R4
+use constantsR4_mod,       only: cp_air, rdgas, grav, rvgas, kappa, pstd_mks, pi
+#else
+use constants_mod,         only: cp_air, rdgas, grav, rvgas, kappa, pstd_mks, pi
+#endif
 use time_manager_mod,       only: time_type, get_time, set_time, operator(+), &
                                   operator(-), operator(/), time_type_to_real
 use fms_mod,                only: error_mesg, FATAL,                 &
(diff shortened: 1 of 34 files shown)


JosephMouallem:
@uramirez8707 @bensonr @laurenchilutti

uramirez8707 (approved):
Looks good to me.

bensonr (approved):
lgtm

I just caution you to check the area of the Earth quantity output during dycore initialization to ensure it doesn't have errors with the OVERLOAD_R4 option and these code modifications.

JosephMouallem:
Thanks for the hint, I will keep that in mind, so far I haven't seen any output error.  Please let me know if there is anything specific to keep an eye on.

lharris4:
@JosephMouallem you should also check the area of the earth in the stdout, especially for both lower- and higher-resolution grids. It has a comparison to the exact value that is very useful for this purpose.

JosephMouallem:
For low resolution 200km cases, the stdout values are exactly the same between the computed global areas (but still slightly different compared to the ideal global area)

I just tried XSHiELD and saw a small difference in the values between 32bit and 64bit:
32bit: GLOBAL AREA (m*m): 0.51009649711388E+15 IDEAL GLOBAL AREA (m*m): 0.51009649655132E+15
64bit: GLOBAL AREA (m*m): 0.51009649711422E+15 IDEAL GLOBAL AREA (m*m): 0.51009649655132E+15 

We can assume this is an acceptable difference?

lharris4:
@JosephMouallem  This is a good result. For reference here is the area values for my original 32-bit X-SHiELD run:

```
GLOBAL AREA (m*m): 0.51009649711428E+15 IDEAL GLOBAL AREA (m*m): 0.51009649655132E+15
```

The difference with your 32 bit run is only at the 12th decimal point, and is smaller than the difference in the computed and true global area in the 64-bit run (9th decimal point). 

The small differences come in from rounding errors in the computation of the angles between grid lengths. Since in FV3 grid cells are bounded by great circles/geodesics, if we know the interior angles and the radius, we can compute the area of the cell exactly. This differs from many dynamical cores that assume grid cells are flat plates, which introduces significant error in computing areas.

More fun with differential geometry: [Wikipedia: Spherical excess](https://en.wikipedia.org/wiki/Spherical_trigonometry#Area_and_spherical_excess) and [Geodesic polygon](https://en.wikipedia.org/wiki/Geodesics_on_an_ellipsoid#Geodesic_polygon)

laurenchilutti (approved):

PR #384, opened 2025-05-23
remove old shield infra, rename shieldfull as shield

JosephMouallem:
This PR removes the old SHiELD infrastructure and renames the new SHiELDFULL configuration, which utilizes the full FMS coupler infrastructure, as SHiELD.

This PR works with:
https://github.com/NOAA-GFDL/SHiELD_build/pull/62
https://github.com/NOAA-GFDL/atmos_drivers/pull/57

### DIFF
--- a/driver/SHiELD/include/atmosphere.inc
+++ b/driver/SHiELD/include/atmosphere.inc
@@ -126,7 +126,6 @@
    select case (index)
 
    case (1)
-#endif
 
 !----------------------
 ! Perform vertical sum:


driver/SHiELD/atmosphere.F90
```diff
@@ -672,15 +721,6 @@ subroutine atmosphere_resolution (i_size, j_size, global)
 
  end subroutine atmosphere_resolution
 
-
```
lharris4, line 675: I am curious as to why these routines needed to be moved to an include file instead of being in `atmosphere.F90` unless it is to prevent that file from becoming too long. Also are these routines the same as in the other drivers? If so, then we might want to move them to a common file instead of having to maintain these routines for all of the different drivers.

driver/SHiELD/atmosphere.F90
```diff
@@ -1048,138 +1062,6 @@ end subroutine atmosphere_nggps_diag
...
- end subroutine get_bottom_wind
-
-
-
- subroutine get_stock_pe(index, value)
-   integer, intent(in) :: index
-   real,   intent(out) :: value
-
-#ifdef USE_STOCK
-   include 'stock.inc'
```
lharris4, line 1133: What is `stock.inc`? I understand it is used mostly by the ESM and likely doesn't concern us.

driver/SHiELD/atmosphere.F90
```diff
@@ -1763,7 +1638,7 @@ subroutine atmos_phys_driver_statein (IPD_Data, Atm_block)
 !!! - "Layer" means "layer mean", ie. the average value in a layer
 !!! - "Level" means "level interface", ie the point values at the top or bottom of a layer
 
-   ptop =  _DBL_(_RL_(Atm(mygrid)%ak(1)))
```
lharris4, line 1766: Glad we can get rid of some of these extra macros.

lharris4 (approved):

driver/SHiELD/atmosphere.F90
bensonr: It allows you to write the basic logic once and then create both 32- and 64-bit versions via macro definitions to have capability of choosing the desired precision at runtime based on the argument type.

lharris4: Cool. Thank you for your explanation.

laurenchilutti (approved):

driver/SHiELD/atmosphere.F90
JosephMouallem: We could remove it. It is taken from the `driver/GFDL/atmosphere.F90` although that file doesn't exist.

lharris4:
If this doesn't exist we should probably remove it.

laurenchilutti (approved):

driver/SHiELD/include/atmosphere.inc
```diff
@@ -120,19 +120,11 @@
...
    real(ATMOSPHERE_KIND_) wm(isc:iec,jsc:jec)
    integer i,j,k
 
    select case (index)
 
-#ifdef USE_STOCK
-   case (ISTOCK_WATER)
-#else
    case (1)
 #endif
```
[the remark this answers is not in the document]
lharris4: Good catch; probably need to remove that `#endif`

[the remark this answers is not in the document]
JosephMouallem: good catch!

PR #385, opened 2025-06-23
Fix pressure variable calculationduring adiabatic initialization `na_init>0`

JosephMouallem:
This PR corrects the calculation of pressure-related variables when adiabatic initialization is enabled.
Previously, inconsistencies between `delp` and the pressure edges `pe` and `peln` lead to incorrect values of `p_bot` and `z_bot`, particularly in the lowest model level. This was especially problematic in coupled configurations, where such inconsistencies caused unphysical values (e.g., `z_bot<0`) and model crashes.


lharris4:
Great, as long as this remains stable it should be kept.

lharris4 (approved):

laurenchilutti (approved):

laurenchilutti:
I will merge when the CI finishes.

PR #395, opened 2025-10-01
Variable gravity 

WChen-NOAA:
**Description**

The model uses the shallow atmosphere approximation with constant gravity
This is problematic for a space weather model (WAM) where Ztop/r0 )~10%
Implement a scaling factor wherever gravity is applied to account for variations with height:

g(r0/r)^2 while setting var_grav = .true. in the name list
the new gravity as grav_var(i,j,k)
if setting var_grav = .false.
The model use an array contains the gravity acceleration constant and the bit of the array follows the compile setting. 
This change will affect the results of non-WAM tests. Previously, the gravity constant was using a mix of 32-bit and 64-bit precision. It has now been changed to 3-D array. it is 32-bit or 64-bit depending on if the 32-bit compilation flag is enabled.

Fixes # ([394](https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/issues/394#issue-3474839215))

**How Has This Been Tested?**

The test results are available vi [doc](https://docs.google.com/presentation/d/1hvlR-eFDRa2Q6b-gPT_drvlrxfM3YtRe/edit?usp=sharing&ouid=101034864378556412510&rtpof=true&sd=true)

**Checklist:**


model/fv_dynamics.F90
```diff
@@ -330,15 +333,39 @@ subroutine fv_dynamics(npx, npy, npz, nq_tot,  ng, bdt, consv_te, fill,
...
       nwat = flagstruct%nwat
       nq = nq_tot - flagstruct%dnats
       nr = nq_tot - flagstruct%dnrts
-      rdg = -rdgas * agrav
+
+      grav_var_h = grav
+      grav_var = grav
+
+      if(flagstruct%var_grav)then
+        do j=js,je
```
lharris4: This loop can be refactored for better locality and cache utilization. I would recommend computing the surface (npz+1) values in the first loop, then compute from npz to 1 by -1 using a  new k-j-i loop. This would also allow you to define `newrad` as a 2D array, reducing memory usage and possibly better fitting in cache.

model/nh_utils.F90
```diff
@@ -2313,20 +2327,27 @@ subroutine nh_BC_k(ptop, grav, kappa, cp, delp, delzBC_t0, delzBC_t1, pt, phis,
...
    rkap = 1./kappa
    peln1 = log(ptop)
-   rdg = - rdgas * rgrav
    denom = 1./BC_split
 
+   do k=1,npz
+     do j=jsd,jed
+       do i=isd,ied
+         rgrav(i,j,k) = 1./grav_var(i,j,k)
+         rdg(i,j,k) = -rdgas/grav_var(i,j,k)
```
lharris4: This line can be replaced with `rdg(i,j,k) = -rdgas*rgrav(i,j,k)` to avoid divisions.

lharris4 (changes requested):
Thank you for submitting this PR. I think this is a nice addition to FV3 and an important first step towards deep atmosphere dynamics. The implementation is done well, mostly self-contained and unobtrusive, and largely follows proper FV3 coding standards. I have a few comments about specific implementation details.

Think carefully about how frequently to update `var_grav`. The Lagrangian vertical coordinate does deform every acoustic timestep, but the variation in gravity `g*(r/r0)**2` even over a `dt_atmos` dynamical timestep will be weak: for an extreme 100 m/s updraft and a 180 s timestep, the elevation change is only 18 km representing a 0.05% change in gravity. My first suggestion would be to simply re-calculate `grav_var` at the beginning of each dynamics timestep and hold it constant through the acoustic and remapping timesteps; this would reduce the amount of re-calculation and data movement, but stability and accuracy would need to be tested.

We also need to verify reproducibility for both hydrostatic and nonhydrostatic cases when `var_grav == .F.`.

yangfanglin:
Lucas, 

Thank you for taking the time to carefully review this PR.   One thing comes to my mind is that wind speed in the thermosphere is normally a few hundred m/s but can reach thousands during geomagnetic storms.  Updating var_grav at each acoustic timestep may become necessary.  @XiaqiongZhou-NOAA @WChen-NOAA 

> Think carefully about how frequently to update `var_grav`. The Lagrangian vertical coordinate does deform every acoustic timestep, but the variation in gravity `g*(r/r0)**2` even over a `dt_atmos` dynamical timestep will be weak: for an extreme 100 m/s updraft and a 180 s timestep, the elevation change is only 18 km representing a 0.05% change in gravity. My first suggestion would be to simply re-calculate `grav_var` at the beginning of each dynamics timestep and hold it constant through the acoustic and remapping timesteps; this would reduce the amount of re-calculation and data movement, but stability and accuracy would need to be tested.

model/dyn_core.F90
```diff
@@ -507,7 +509,24 @@ subroutine dyn_core(npx, npy, npz, ng, sphum, nq, bdt, n_map, n_split, zvir, cp,
      if ( it==1 ) then
                                        call timing_on('COMM_TOTAL')
           call complete_group_halo_update(i_pack(1), domain)
+          if(flagstruct%var_grav) call complete_group_halo_update(i_pack(2), domain)
                                       call timing_off('COMM_TOTAL')
+          do k=1,npz
+            do j=jsd,jed
+              do i=isd,ied
+                grav_var(i,j,k) = (grav_var_h(i,j,k+1)+grav_var_h(i,j,k))/2.
```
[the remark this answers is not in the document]
WChen-NOAA: by the way, the grav_var_h updated at each remapping loop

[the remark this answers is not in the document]
lharris4: OK. I presume this is just to remap grav_var at layer interfaces onto layer centroids, done once per remapping loop.

model/fv_dynamics.F90
```diff
@@ -769,14 +798,31 @@ subroutine fv_dynamics(npx, npy, npz, nq_tot,  ng, bdt, consv_te, fill,
...
+                                                  call timing_off('Remapping')
 
      if ( flagstruct%molecular_diffusion ) then
 ! do thermosphere adjustment if it is turned on and at GFDL_interstitial%last_step.
          if( md_tadj_layers .gt.0 .and. md_time .and. GFDL_interstitial%last_step ) then
              call thermosphere_adjustment(domain,gridstruct,npz,bd,ng,pt)
         endif ! md_tadj_layers>0 and md_time and GFDL_interstitial%last_step
      endif
+      
+     if(flagstruct%var_grav)then
```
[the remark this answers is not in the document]
lharris4: After some discussion, updating on the remapping timestep is probably the best.

laurenchilutti:
@WChen-NOAA Do you anticipate any changes to this PR? Is this only waiting on review?

XiaqiongZhou-NOAA:
Wen will likely make adjustments to ensure the code reproduces w/o this PR.

XiaqiongZhou-NOAA (approved):
It seems fine to me.

XiaqiongZhou-NOAA:
Wen has resolved all the issues, and everything appears to be working as expected. @lharris4, do you have any additional feedback or suggestions before we merge it to ufs-model?

lharris4 (approved):
I think this is good. In the future I would like to see `MULTI_GASES` cleaned up to be a runtime switch and not a compiler directive, similar to what we did for `USE_COND` and `MOIST_CAPPA` in main.

gspetro-NOAA:
@vithikashah001 We will probably process parent PR https://github.com/ufs-community/ufs-weather-model/pull/2916 today or early next week. Do you think this needs to be reviewed by someone with write access? I think lately we haven't worried about that, as long as there are appropriate reviewers, but one of my colleagues was asking, so I just wanted to check on the policy for this repo.

model/dyn_core.F90
```diff
@@ -3152,7 +3174,7 @@ subroutine molecular_diffusion_run(u,v,w,delp,pt,pkz,cappa,q,bd,   &
   type(fv_grid_type),  intent(INOUT), target :: gridstruct
   type(fv_flags_type), intent(IN),    target :: flagstruct
   type(domain2d), intent(inout) :: domain
-  type(group_halo_update_type), intent(inout) :: i_pack(*)
+  type(group_halo_update_type):: i_pack(7)
```
JosephMouallem, line 3177: Is there a reason why `i_pack` is made local with different indices in this routine?
One suggestion is to use the `complete=.false.` for these successive communication calls with `complete=.true.` for the last call for better performance.

model/nh_core.F90
```diff
@@ -105,16 +107,25 @@ subroutine Riem_Solver3(ms, dt,   is,   ie,   js, je, km, ng,    &
...
   real, dimension(is:ie,km):: kapad2
 #endif
-  real gama, rgrav, ptk, peln1
+  real, dimension(isd:ied,jsd:jed,km) :: rgrav
+  real gama, ptk, peln1
   integer i, j, k
 
     gama = 1./(1.-akap)
-   rgrav = 1./grav
+    do k=1,km
```
JosephMouallem, line 117: An OpenMP directive here would be a nice improvement

model/nh_utils.F90
```diff
@@ -379,17 +380,26 @@ subroutine Riem_Solver_c(ms,   dt,  is,   ie,   js, je, km,   ng,  &
...
   integer is1, ie1
 
     gama = 1./(1.-akap)
-   rgrav = 1./grav
+
 
    is1 = is - 1
    ie1 = ie + 1
 
+   do k=1,km
```
JosephMouallem, line 396: Same here, an OpenMP directive here would be a nice improvement

JosephMouallem (approved):

model/nh_core.F90
WChen-NOAA: Thank you for suggestions. The Open MP here is added.

model/nh_utils.F90
WChen-NOAA: Added. And thanks again

model/dyn_core.F90
WChen-NOAA: Thank you very much for your suggestions. 
We added it for the WAM implementation.  When we checking the WAM results, the values in halo area are not updated properly and after many works we think the problem could be happened in this routine. And now the WAM looks fine and I would like to keep this. And This only affect WAM tests.

FernandoAndrade-NOAA:
@bensonr @vithikashah001 There are 3 review approvals on this PR, however none are from users with write access to the repo. It would be preferable to have an approval from a repo maintainer as well in preparation for processing of the related Weather Model PR, please review/comment/approve this PR.

bensonr:
> @bensonr @vithikashah001 There are 3 review approvals on this PR, however none are from users with write access to the repo. It would be preferable to have an approval from a repo maintainer as well in preparation for processing of the related Weather Model PR, please review/comment/approve this PR.

@FernandoAndrade-NOAA - as the science team has okayed the PR and the EPIC team has confirmed it's efficacy, no actual review is needed by the repo management team.

gspetro-NOAA:
Testing completed successfully for https://github.com/ufs-community/ufs-weather-model/pull/2916. This PR can be merged.

PR #400, opened 2026-01-07
fully integrated physics

linjiongzhou:
**Description**

This PR is based on the latest main branch and includes the fully integrated physics framework developed since the start of this effort. When running with the original physics configuration, results should remain unchanged.

To enable the integrated physics, add the following namelist:

&integ_phys_nml
  do_fast_phys   = .T.
  do_inline_pbl  = .T.
  do_inline_cnv  = .T.
  do_inline_gwd  = .T.
  inline_cnv_flag = 2
  inline_pbl_flag = 1
/

&sa_sas_nml
/

&sa_aamf_nml
/

&sa_tke_edmf_nml
/

&sa_tke_edmf_new_nml
/

&sa_gwd_nml
/

Fixes # (issue)

**How Has This Been Tested?**

This functionality has been exercised extensively throughout the development process.

**Checklist:**

Please check all whether they apply or not
- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [X] Any dependent changes have been merged and published in downstream modules

PR #84, opened 2021-03-25
mpi_maxval and mpi_minval reductions in fv_mp_mod (PR to master)

SamuelTrahanNOAA:
A PR to FV3 to add per-timestep pgr diagnostics needs an `mpi_reduce_maxval_r8`. These changes add that function, as well as `minval` and `_r4` varieties for completeness.

Relevant FV3 PR:

https://github.com/NOAA-GSL/fv3atm/pull/78

PR #413, opened 2026-04-24
add integrated physics

JosephMouallem:
Add integrated physics from the original PR https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/400
credit to @linjiongzhou 

Dependencies:
https://github.com/NOAA-GFDL/atmos_drivers/pull/72
https://github.com/NOAA-GFDL/SHiELD_physics/pull/84

### DIFF
--- a/driver/SHiELD/atmosphere.F90
+++ b/driver/SHiELD/atmosphere.F90
@@ -90,6 +90,11 @@ use fv_grid_utils_mod,  only: g_sum
 
 use mpp_domains_mod, only:  mpp_get_data_domain, mpp_get_compute_domain
 use gfdl_mp_mod,        only: gfdl_mp_init, gfdl_mp_end
+use sa_tke_edmf_mod,    only: sa_tke_edmf_init
+use sa_tke_edmf_new_mod,only: sa_tke_edmf_new_init
+use sa_sas_mod,         only: sa_sas_init
+use sa_aamf_mod,        only: sa_aamf_init
+use sa_gwd_mod,         only: sa_gwd_init
 use diag_manager_mod,   only: send_data
 use external_aero_mod,  only: load_aero, read_aero, clean_aero
 use coarse_graining_mod, only: coarse_graining_init
@@ -331,6 +336,15 @@ contains
    allocate(pref(npz+1,2), dum1d(npz+1))
 
    call gfdl_mp_init(input_nml_file, stdlog(), Atm(mygrid)%flagstruct%hydrostatic)
+   if (Atm(mygrid)%flagstruct%do_inline_pbl) then
+     if (Atm(mygrid)%flagstruct%inline_pbl_flag .eq. 1) call sa_tke_edmf_init(input_nml_file, stdlog())
+     if (Atm(mygrid)%flagstruct%inline_pbl_flag .eq. 2) call sa_tke_edmf_new_init(input_nml_file, stdlog())
+   endif
+   if (Atm(mygrid)%flagstruct%do_inline_cnv) then
+     if (Atm(mygrid)%flagstruct%inline_cnv_flag .eq. 1) call sa_sas_init(input_nml_file, stdlog())
+     if (Atm(mygrid)%flagstruct%inline_cnv_flag .eq. 2) call sa_aamf_init(input_nml_file, stdlog())
+   endif
+   if (Atm(mygrid)%flagstruct%do_inline_gwd) call sa_gwd_init(input_nml_file, stdlog())
 
    call timing_on('FV_RESTART')
    call fv_restart(Atm(mygrid)%domain, Atm, dt_atmos, seconds, days, cold_start, &
@@ -545,6 +559,7 @@ if ( is_master() ) write(*,*) 'CALL atmos_global_diag_init'
                       Atm(n)%flagstruct, Atm(n)%neststruct,                &
                       Atm(n)%thermostruct, Atm(n)%idiag, Atm(n)%bd,        &
                       Atm(n)%parent_grid, Atm(n)%domain, Atm(n)%inline_mp, &
+                      Atm(n)%inline_pbl, Atm(n)%inline_cnv, Atm(n)%inline_gwd, &
                       Atm(n)%heat_source,Atm(n)%diss_est,time_total=time_total)
 
      call timing_off('FV_DYNAMICS')
@@ -722,11 +737,11 @@ if ( is_master() ) write(*,*) 'CALL atmos_global_diag_init'
  end subroutine atmosphere_resolution
 
  subroutine atmosphere_control_data (i1, i2, j1, j2, kt, p_hydro, hydro, tile_num, &
-                                     do_inline_mp, do_cosp)
+                                     do_inline_mp, do_inline_cnv, do_inline_pbl, do_inline_gwd, do_cosp)
    integer, intent(out)           :: i1, i2, j1, j2, kt
    logical, intent(out), optional :: p_hydro, hydro
    integer, intent(out), optional :: tile_num
-   logical, intent(out), optional :: do_inline_mp, do_cosp
+   logical, intent(out), optional :: do_inline_mp, do_inline_cnv, do_inline_pbl, do_inline_gwd, do_cosp
    i1 = Atm(mygrid)%bd%isc
    i2 = Atm(mygrid)%bd%iec
    j1 = Atm(mygrid)%bd%jsc
@@ -737,6 +752,9 @@ if ( is_master() ) write(*,*) 'CALL atmos_global_diag_init'
    if (present(  hydro))   hydro = Atm(mygrid)%flagstruct%hydrostatic
    if (present(tile_num)) tile_num = Atm(mygrid)%global_tile
    if (present(do_inline_mp)) do_inline_mp = Atm(mygrid)%flagstruct%do_inline_mp
+   if (present(do_inline_cnv)) do_inline_cnv = Atm(mygrid)%flagstruct%do_inline_cnv
+   if (present(do_inline_pbl)) do_inline_pbl = Atm(mygrid)%flagstruct%do_inline_pbl
+   if (present(do_inline_gwd)) do_inline_gwd = Atm(mygrid)%flagstruct%do_inline_gwd
    if (present(do_cosp)) do_cosp = Atm(mygrid)%flagstruct%do_cosp
 
  end subroutine atmosphere_control_data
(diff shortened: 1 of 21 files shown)


tools/fv_restart.F90
```diff
@@ -377,7 +377,8 @@ subroutine fv_restart(fv_domain, Atm, dt_atmos, seconds, days, cold_start, grid_
                      Atm(n)%flagstruct%dry_mass, Atm(n)%flagstruct%mountain, &
                      Atm(n)%flagstruct%moist_phys, Atm(n)%flagstruct%hydrostatic, &
                      hybrid, Atm(n)%delz, Atm(n)%ze0, Atm(n)%ks, Atm(n)%ptop, &
-                     Atm(n)%domain, Atm(n)%tile_of_mosaic, Atm(n)%bd)
+                     Atm(n)%domain, Atm(n)%tile_of_mosaic, Atm(n)%bd, &
+                     Atm(n)%inline_pbl, Atm(n)%inline_gwd)
```
lharris4, line 381: Are these actually used in this routine?

gaokun227:
I think it will be helpful to indicate in the description if the code changes in this PR are identical to those in Linjiong's original PR https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/pull/400.

driver/SHiELD/atmosphere.F90
```diff
@@ -331,6 +336,15 @@ subroutine atmosphere_init (Time_init, Time, Time_step, Grid_box, area, IAU_Data
    allocate(pref(npz+1,2), dum1d(npz+1))
 
    call gfdl_mp_init(input_nml_file, stdlog(), Atm(mygrid)%flagstruct%hydrostatic)
+   if (Atm(mygrid)%flagstruct%do_inline_pbl) then
```
lharris4, line 339: Eventually these should be moved into a separate integrated physics initialization routine.

driver/SHiELD/atmosphere.F90
```diff
@@ -1200,9 +1218,46 @@ subroutine atmosphere_state_update (Time, IPD_Data, IAU_Data, Atm_block)
          Atm(n)%delp(i,j,k1) = q0
          Atm(n)%q(i,j,k1,1:nq_adv) = qwat(1:nq_adv) / q0
          if (dnats .gt. 0) Atm(n)%q(i,j,k1,nq_adv+1:nq) = IPD_Data(nb)%Stateout%gq0(ix,k,nq_adv+1:nq)
+         if (Atm(n)%flagstruct%do_inline_pbl) Atm(n)%inline_pbl%radh(i,j,k1) = IPD_Data(nb)%Stateout%radh(ix,k)
```
lharris4, line 1221: Is this radiative heating?

model/fast_phys.F90
```diff
@@ -184,6 +200,960 @@ subroutine fast_phys (is, ie, js, je, isd, ied, jsd, jed, km, npx, npy, nq, &
...
+                    !inline_pbl%dusfc (is:ie, j), inline_pbl%dvsfc (is:ie, j), &
+                    !inline_pbl%dtsfc (is:ie, j), inline_pbl%dqsfc (is:ie, j))
+
+            endif
+
+            ! update u, v, T, q, and delp, vertical index flip over
+            do k = 1, km
+                kr = km - k + 1
+                q (is:ie, j, kr, nwat+1:nq) = qa (is:ie, k, nwat+1:nq)
+                dqv = qa (is:ie, k, sphum) - q (is:ie, j, kr, sphum)
```
lharris4, line 517: If we independently develop our own PBL scheme, eventually it should be brought up to use GFDL standards for handling arbitrary tracers.

model/fast_phys.F90
```diff
@@ -184,6 +200,960 @@ subroutine fast_phys (is, ie, js, je, isd, ied, jsd, jed, km, npx, npy, nq, &
...
+                q_liq = q (is:ie, j, kr, liq_wat) + q (is:ie, j, kr, rainwat)
+                q_sol = q (is:ie, j, kr, ice_wat) + q (is:ie, j, kr, snowwat) + q (is:ie, j, kr, graupel)
+                ta (is:ie, k) = pt (is:ie, j, kr) / ((1. + r_vir * q (is:ie, j, kr, sphum)) * &
+                    (1. - (q_liq + q_sol)))
+                uu (is:ie, k) = ua (is:ie, j, kr)
+                vv (is:ie, k) = va (is:ie, j, kr)
+            enddo
+
+            ! check if pressure or height cross over
+            if (safety_check) then
```
lharris4, line 932: This is a nice feature to avoid instabilities.

model/intermediate_phys.F90
```diff
@@ -462,102 +528,66 @@ subroutine intermediate_phys (is, ie, js, je, isd, ied, jsd, jed, km, npx, npy,
...
-            if (allocated (inline_mp%qg_dt)) inline_mp%qg_dt (is:ie, j, kmp:km) = &
-                inline_mp%qg_dt (is:ie, j, kmp:km) - q (is:ie, j, kmp:km, graupel)
-            if (allocated (inline_mp%t_dt)) inline_mp%t_dt (is:ie, j, kmp:km) = &
-                inline_mp%t_dt (is:ie, j, kmp:km) - pt (is:ie, j, kmp:km)
-            if (allocated (inline_mp%u_dt)) inline_mp%u_dt (is:ie, j, kmp:km) = &
-                inline_mp%u_dt (is:ie, j, kmp:km) - ua (is:ie, j, kmp:km)
-            if (allocated (inline_mp%v_dt)) inline_mp%v_dt (is:ie, j, kmp:km) = &
-                inline_mp%v_dt (is:ie, j, kmp:km) - va (is:ie, j, kmp:km)
+            kinver = km
+            lsoil = 4
```
lharris4, line 558: Note this is only strictly true for NOAH LSM and Noah-MP. However I don't know the purpose of lsoil here. I think it is incorrect as it is only used for `sfc_seai()`, which is the energy balance scheme for the 3-level ice model, although in a pinch it may provide a useful surrogate for the surface variables over land necessary to run the PBL scheme.

model/intermediate_phys.F90
```diff
@@ -570,158 +600,1673 @@ subroutine intermediate_phys (is, ie, js, je, isd, ied, jsd, jed, km, npx, npy,
...
-               if (thermostruct%moist_kappa) then
-                  pkz (is:ie, j, kmp:km) = exp (cappa (is:ie, j, kmp:km) * &
-                       log (rrg * delp (is:ie, j, kmp:km) / &
-                       delz (is:ie, j, kmp:km) * pt (is:ie, j, kmp:km)))
-               else
-                  pkz (is:ie, j, kmp:km) = exp (akap * log (rrg * delp (is:ie, j, kmp:km) / &
-                       delz (is:ie, j, kmp:km) * pt (is:ie, j, kmp:km)))
-               endif
+                if (thermostruct%moist_kappa) then
+                    pkz (is:ie, j, 1:km) = exp (cappa (is:ie, j, 1:km) * &
```
lharris4, line 832: This is an important bugfix. `kmp` should only be used for microphysics, as there is a layer above which microphysical processes may be very slow or nonexistent; but the other physics schemes (both PBL and GWD) are still active above `k=kmp`.

model/intermediate_phys.F90
```diff
@@ -570,158 +600,1673 @@ subroutine intermediate_phys (is, ie, js, je, isd, ied, jsd, jed, km, npx, npy,
...
+
+        ! update D grid wind
+        call update_dwinds_phys (is, ie, js, je, isd, ied, jsd, jed, abs (mdt), u_dt, v_dt, u, v, &
+                 gridstruct, npx, npy, km, domain)
+
+        deallocate (u_dt)
+        deallocate (v_dt)
+
+        ! update dry total energy
+        if (consv .gt. consv_min) then
```
lharris4, line 1907: This energy checker code can go into a subroutine since it is repeated several times in the intermediate and fast physics drivers.

lharris4 (approved):
I think this is good. I have left a number of comments about stylistic matters, but I couldn't find any bugs that need to be fixed before merging this in. A few general comments:

- The integrated PBL scheme may be very difficult to get to work with implicitly-coupled land as we would then need to break up `fv_dynamics.F90` into up and down calls, and always use k_split = 1. But if we constrain the integrated physics to only work with explicit coupled land, then we are OK. I do see that this PR includes a lightweight version of the surface energy balance scheme (taken from the simple 3-layer ice scheme) to provide surface inputs for the PBL scheme; in the future we may need to take care that this is not inconsistent with the surface schemes used in the rest of the model.

- There is also a lot of 'index flipping' necessary to use the k-ordering (k=1 is the bottom) from the SHiELD physics, which may be slowing down the model. In the future, we should re-write the integrated physics schemes to use FV3's k-ordering (k=1 is the top), which is used for all AMx physics packages. This will be necessary anyway as we move towards unifying the AM5 and SHiELD physics.

- Do we have tests for all of the inline physical parameterizations? If so, they would be useful to add to the CI/CD.

lharris4:
> I do see that this PR includes a lightweight version of the surface energy balance scheme (taken from the simple 3-layer ice scheme) to provide surface inputs for the PBL scheme; in the future we may need to take care that this is not inconsistent with the surface schemes used in the rest of the model.

After reading 
[SHiELD_Physics #84](https://github.com/NOAA-GFDL/SHiELD_physics/pull/84/changes) I now understand better that the ocean and ice schemes are now fully in-lined with the PBL scheme, which is a good move since (a) they are very cheap and (b) we never need to worry about implicit coupling for these, once we move to fully coupled ice and ocean models.

gaokun227:
> > I do see that this PR includes a lightweight version of the surface energy balance scheme (taken from the simple 3-layer ice scheme) to provide surface inputs for the PBL scheme; in the future we may need to take care that this is not inconsistent with the surface schemes used in the rest of the model.
> 
> After reading [SHiELD_Physics #84](https://github.com/NOAA-GFDL/SHiELD_physics/pull/84/changes) I now understand better that the ocean and ice schemes are now fully in-lined with the PBL scheme, which is a good move since (a) they are very cheap and (b) we never need to worry about implicit coupling for these, once we move to fully coupled ice and ocean models.

This is consistent with my understanding too. We should not run inline PBL in the fully coupled mode.

gaokun227 (approved):
This PR seems consistent with Linjiong's original PR and it looks good to me. I will add some of my modifications in a future PR for the 3DTKE scheme.

lharris4:
Hi, all. Is this ready to merge?

tools/fv_restart.F90
JosephMouallem: When do `do_inline_pbl` or `do_inline_gwd` are turned on, several variables under `inline_pbl` and `inline_gwd` are set. Not sure if there a cleaner way to do this.

driver/SHiELD/atmosphere.F90
JosephMouallem: It looks it is from the definition provided in the physics https://github.com/NOAA-GFDL/SHiELD_physics/pull/84

model/intermediate_phys.F90
JosephMouallem: `kmp` has been changed in different places not only here, we might need to take a closer look at it to make sure the code is working as intended.

JosephMouallem:
> Hi, all. Is this ready to merge?

There might be some subtle issues with this code that were found during testing, Kun and I are looking into them. Sorry for the delayed update.

JosephMouallem:
The integrated physics PRs are ready to merge. These changes do not affect the answers for either the uncoupled or coupled models when the new options are disabled, and they work as intended when enabled. I'm okay with merging.

laurenchilutti:
The Ci is back up and running (thank you Vithika!!) but this PR is failing due to a seg fault.  Check it out here: https://github.com/NOAA-GFDL/GFDL_atmos_cubed_sphere/actions/runs/30284920393/job/91737236529?pr=413

laurenchilutti (approved):
Looks like this works now! I feel comfortable with this and the associated PRs being merged.  Just be sure to merge https://github.com/NOAA-GFDL/SHiELD_physics/pull/84 and https://github.com/NOAA-GFDL/atmos_drivers/pull/72 first, then take out the changes to `.github/workflows/SHiELD_parallelworks_intel.yaml`, wait for the CI to run again (it will pass), then merge this PR

PR #419, opened 2026-07-09
Updates for ccpp-capgen v1

dustinswales:
**Description**
This PR contains the necessary changes to the CCPP fast-physics interfaces for the new code generator capgen v1. 

**How Has This Been Tested?**
All UFS weather model regression tests pass and are bit-for-bit identical.

** Dependencies
- NCAR/ccpp-framework#762
- ufs-community/ccpp-physics#378
- NOAA-EMC/ufsatm#1125
- NOAA-EMC/CMEPS#173
- ufs-community/ufs-weather-model#3288

### DIFF
--- a/driver/UFS/atmosphere.F90
+++ b/driver/UFS/atmosphere.F90
@@ -498,12 +498,6 @@ contains
 
    ! Do CCPP fast physics initialization before call to adiabatic_init (since this calls fv_dynamics)
 
-   ! For fast physics running over the entire domain, block
-   ! and thread number are not used; set to safe values
-   !cdata%blk_no = 1
-   !cdata%thrd_no = 1
-   !cdata%thrd_cnt = 1
-
    ! Create shared data type for fast and slow physics, one for each thread
 #ifdef OPENMP
    nthreads = omp_get_max_threads()
--- a/driver/UFS/atmosphere.F90
+++ b/driver/UFS/atmosphere.F90
@@ -297,10 +297,9 @@ contains
 !! and diagnostics.
  subroutine atmosphere_init (Time_init, Time, Time_step, Grid_box, area)
 
-   use ccpp_static_api,   only: ccpp_physics_init
-   use CCPP_data,         only: ccpp_suite,          &
-                                cdata => cdata_tile, &
-                                GFDL_interstitial
+   use ufs_ccpp_cap,      only: ccpp_physics_init
+   use CCPP_driver,       only: ccpp_suite, errmsg, errflg
+   use CCPP_data,         only: GFDL_interstitial
 #ifdef OPENMP
    use omp_lib
 #endif
@@ -501,9 +500,9 @@ contains
 
    ! For fast physics running over the entire domain, block
    ! and thread number are not used; set to safe values
-   cdata%blk_no = 1
-   cdata%thrd_no = 1
-   cdata%thrd_cnt = 1
+   !cdata%blk_no = 1
+   !cdata%thrd_no = 1
+   !cdata%thrd_cnt = 1
 
    ! Create shared data type for fast and slow physics, one for each thread
 #ifdef OPENMP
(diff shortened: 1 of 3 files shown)


driver/UFS/atmosphere.F90
```diff
@@ -501,9 +500,9 @@ subroutine atmosphere_init (Time_init, Time, Time_step, Grid_box, area)
 
    ! For fast physics running over the entire domain, block
    ! and thread number are not used; set to safe values
-   cdata%blk_no = 1
-   cdata%thrd_no = 1
-   cdata%thrd_cnt = 1
+   !cdata%blk_no = 1
```
climbfuji: Remove lines 501-505?

lharris4: These really should be deleted.

model/fv_dynamics.F90
```diff
@@ -198,10 +198,9 @@ subroutine fv_dynamics(npx, npy, npz, nq_tot,  ng, bdt, consv_te, fill,
                         parent_grid, domain, diss_est, inline_mp)
 
     use mpp_mod,           only: FATAL, mpp_error
-    use ccpp_static_api,   only: ccpp_physics_timestep_init,    &
-                                 ccpp_physics_timestep_finalize
-    use CCPP_data,         only: ccpp_suite
-    use CCPP_data,         only: cdata => cdata_tile
+    use ufs_ccpp_cap,      only: ccpp_physics_timestep_init,    &
```
lharris4, line 201: Do the CCPP routines need to be in `fv_dynamics.F90`? I would recommend moving all into CCPP routines into `fv_mapz.F90` as part of the intermediate physics driver. The next FV3 release will include the intermediate and fast physics drivers, which can then more cleanly handle implementation-specific interfaces.

lharris4: Note that `Eulerian_to_Lagrangian()` in `fv_mapz.F90` already has an argument `last_step` so you can see when you need to call the `_final` routine.

lharris4 (changes requested):

model/fv_dynamics.F90
dustinswales: @lharris4 I'm terribly familiar with the dynamics to physics coupling in the FV3 UFS, so iI don't why it's like this.
ccpp_timestep_init and ccpp_timestep_finalize use the full domain, whereas ccpp_physics_run is run on pieces of the full domain. So maybe that is why? @climbfuji Might have some insight on this.

In any event, as part of this PR we are just replacing the API's to the physics. We can always open an issue and revisit this if need be.

climbfuji: I can take a look later. First, I need to remind myself why it is this way. I believe it's because the init/final steps need to run in different places that are not in fv_mapz.F90.

climbfuji: I agree with @dustinswales that this PR should not attempt to change when and where specific CCPP routines are called. We want to minimize the code changes across the UFS to what is needed for transitioning from ccpp-prebuild to ccpp-capgen-v1.

But I also agree with @lharris4 that we should look into consolidating the CCPP interfaces into as few as possible places. Just not in this PR, but in a follow-up PR.

For this, we need to be able to run the `ccpp_register`, `ccpp_init`, and `ccpp_physics_init` steps once at the right time and with access to the necessary information, and similar for the timestep/run parts and the `_final` parts. It will be easier to do this if we let the CCPP framework instead of the FV3 dycore handle the CCPP interstitial data. This is another argument why we want to do this in a separate PR.

lharris4: OK. How does NCAR implement CCPP timestepping in their weather model? Perhaps that would give us an idea of how to proceed here.

climbfuji: > OK. How does NCAR implement CCPP timestepping in their weather model? Perhaps that would give us an idea of how to proceed here.

Lucas, happy to meet with you and walk you through the steps. Both Dustin and I will be traveling the next 2-3 weeks. After that if you're ok?

lharris4: Thank you, but I don't need this level of detail. My principal concern is ensuring that FV3's physics drivers and external dependencies remain well-organized.

dustinswales:
@lharris4 Are you okay with these existing changes? If so, can you approve this PR.
I will open an issue on cleaning up the FV3-CCPP integration so we don't forget.

lharris4:
> @lharris4 Are you okay with these existing changes? If so, can you approve this PR. I will open an issue on cleaning up the FV3-CCPP integration so we don't forget.

Sure, please do open this issue. I will approve the PR.

lharris4 (approved):

gspetro-NOAA:
> @lharris4 Are you okay with these existing changes? If so, can you approve this PR. I will open an issue on cleaning up the FV3-CCPP integration so we don't forget.

@dustinswales Could you open the issue before we process your PR so that we don't accidentally lose track?

dustinswales:
> > @lharris4 Are you okay with these existing changes? If so, can you approve this PR. I will open an issue on cleaning up the FV3-CCPP integration so we don't forget.
> 
> @dustinswales Could you open the issue before we process your PR so that we don't accidentally lose track?

See #421

gspetro-NOAA:
> > > @lharris4 Are you okay with these existing changes? If so, can you approve this PR. I will open an issue on cleaning up the FV3-CCPP integration so we don't forget.
> > 
> > 
> > @dustinswales Could you open the issue before we process your PR so that we don't accidentally lose track?
> 
> See #421

Awesome, thanks! 🙂

gspetro-NOAA:
@vithikashah001 Does anyone else need to review this PR before we proceed with testing at the WM level?

gspetro-NOAA:
Testing for https://github.com/ufs-community/ufs-weather-model/pull/3288 completed successfully. This PR can be merged.

JosephMouallem (approved):

PR #426, opened 2026-09-18
fix: use if-elseif logic for choosing the inline PBL scheme option

gaokun227:
This PR fixes a bug in the conditional logic for selecting the inline PBL scheme.

Previously, the single-line `if` statements only applied to the surface-layer calls, leaving the main PBL calls outside the conditional statements. As a result, both `sa_tke_edmf_pbl` and `sa_tke_edmf_new_pbl` were executed regardless of the value of `inline_pbl_flag`.

This PR replaces the single-line conditions with an `if/elseif` block so that only the PBL scheme selected by `inline_pbl_flag` is executed.

This bug fix changes answer if inline PBL is activated.


lharris4 (approved):
Those continuation characters can cause trouble! Thanks for catching this.

model/intermediate_phys.F90
```diff
@@ -758,6 +758,7 @@ subroutine intermediate_phys (is, ie, js, je, isd, ied, jsd, jed, km, npx, npy,
                     inline_pbl%hpbl (is:ie, j), inline_pbl%kpbl (is:ie, j))
                     !inline_pbl%dusfc (is:ie, j), inline_pbl%dvsfc (is:ie, j), &
                     !inline_pbl%dtsfc (is:ie, j), inline_pbl%dqsfc (is:ie, j))
+            endif
```
lharris4: Do you wish to put in a warning or error flag if `inline_pbl_flag` is an invalid value?

JosephMouallem (approved):

gaokun227:
@vithikashah001 Could you please merge this PR? It will set the stage for a larger PR that I plan to submit later this week. Thank you.

